PR #2, opened 2019-03-27
Make release notes checks work with .po and .pot files

loremattei:
I noticed that the release notes check fails on the WooCommerce apps because they use .pot files instead of .po. 

I'm also not sure how we can test it.


loremattei:
Thanks for looking James!

PR #4, opened 2019-05-31
Add Simplenote repositories

jtreanor:
I have enabled Peril for Simplenote-iOS, Simplenote-macOS and Simplenote-Android. They are currently just using the default rules (from [common.ts](https://github.com/wordpress-mobile/peril-settings/blob/master/org/common.ts)).

This updates the config so the use the appropriate platform-specific rules.

Have a look through [ios-macos.ts](https://github.com/wordpress-mobile/peril-settings/blob/17783795d7d2b5a16df8744ee591f048c75adcc8/org/ios-macos.ts) and [android.ts](https://github.com/wordpress-mobile/peril-settings/blob/17783795d7d2b5a16df8744ee591f048c75adcc8/org/android.ts) to confirm that these are the correct rules to be using for the repos.


jtreanor:
@theck13 These rules are shared with the WooCommerce, WordPress and now Simplenote mobile projects so that is why it’s here. We could have a separate repo per org if we thought that was useful.

Yeah, that’s the only Android rule right now. We will likely add more in future.

jtreanor:
As requested by @theck13, I have updated this PR to refactor our settings so that only the label rules and the iOS/Android rules are applied.

@theck13 and @bummytime can you take another look?

All other repos are unchanged.

jtreanor:
Thanks folks!

The new settings should take effect now. Let me know if you have any issues.

PR #5, opened 2019-06-12
Automatically create merge branch when PR contains subtree changes

markpar:
<< Coming soon :) >>


org/prForSubtreeChanges.ts
```diff
@@ -0,0 +1,132 @@
...
+            for (let file of deletedFiles) {
+                console.log(`Processing deleted file ${file}`);
+                console.log("Getting SHA");
+                let fileSha = (await api.repos.getContent({owner: org, repo: destRepo, path: file})).data.sha;
+                console.log(`Got ${destRepo}/${file}: SHA ${fileSha}`);
+
+                console.log(`Deleting ${file} in ${mergeBranch}`);
+                let commitSha = (await api.repos.deleteFile({owner: org, repo: destRepo, path: file, sha: fileSha, branch: mergeBranch, message: commitMessage})).data.commit.sha;
+                console.log(`Deleted ${file}: SHA ${commitSha}`)
+            }
```
jtreanor: Thanks for sharing the in progress PR @markpar!

I'm not sure mirroring created, added and deleted files from the PR is the same as a subtree push. For example, if `libs/login` already had changes not in `WordPress-Login-Flow-Android`, this code would only update the files changed in the new PR rather than bringing it fully up to date.

Of course that would not be a problem if the created `WordPress-Login-Flow-Android` PR is always merged before the WP/Woo PR, but its worth considering if we are happy with this behaviour.

Also, I'm not very familiar with `git-subtree`'s internals. Might be worth checking if applying changes like this would be the same as what a `git subtree push` does.

jtreanor:
I realise that this is a draft but it may be worth considering adding some tests for these changes. Its quite a bit more complex than any of our existing rules. You can take a look at Artsy's Peril settings [here](https://github.com/artsy/peril-settings) to see how they do that.

org/prForSubtreeChanges.ts
markpar: > I'm not sure mirroring created, added and deleted files from the PR is the same as a subtree push.

Good catch, and you are correct that they are not exactly the same thing.  There's not a lot of documentation on `git-subtree` but I did find the source for it: https://github.com/git/git/blob/master/contrib/subtree/git-subtree.sh.  A subtree push does the equivalent of creating a temporary subtree and then merge, so I think this approach does work, though not as nicely as `subtree push`.  

>if libs/login already had changes not in WordPress-Login-Flow-Android, this code would only update the files changed in the new PR

That's correct and is currently a limitation.  The thinking here is that when someone makes changes in `WordPress-Android/libs/login/`, a merge branch back into `WordPress-Login-Flow-Android` will be created with the changes and that the PR author is responsible for completing the merge.  Doing a full-on `subtree push` would ensure that no changes slip through the cracks, but I've not yet found a way to do that on the Heroku server.

markpar:
Closing this PR in favour of a simpler initial approach, https://github.com/Automattic/peril-settings/pull/6.

PR #6, opened 2019-06-14
Add subtree push instructions

markpar:
This change is the initial implementation of a feature request by @aforcier to automate the process of synchronising changes between the `WordPress-Android` subtree `libs/login/` and its source repo, `WordPress-Login-Flow-Android`.  

Currently the process is done manually, and by one person.  This change adds a rule that runs when a PR is opened, scanning it for any changes in `libs/login/`.  If it finds any, a warning is written to the PR with step-by-step instructions for creating a PR to merge those changes back into `WordPress-Login-Flow-Android`.

I haven't tested this across forks and currently the org/owner name is hardcoded to `Automattic`.  This plus two other hardcoded values, the destination repo name and subtree path to monitor, might be discoverable at runtime (well, not the subtree path, but maybe the other two 😃 ), enabling this rule to be used for any subtree.  

Further improvements to this rule could include actually creating the pull request and monitoring it for merge conflicts.

Currently, the rule is enabled only for PRs opened in the `WordPress-Android` repo.  Once we are confident in the rule, we can enable it for `woocommerce-android`, as well.

Major thanks for @jtreanor for amazing assistance and feedback!

Here's a screenshot of what the instructions look like.  Paths are relative to my forks.
image


org/github/pr-for-subtree.ts
```diff
@@ -0,0 +1,46 @@
...
+        // The name of the branch that the `subtree push` command will create.
+        const mergeBranch = `merge/${pr.repo}/${pr.number}`;
+
+        // The merge instructions.
+        let markdownText: string;
+
+        // Put it all together! 
+        markdownText = `This PR contains changes in the subtree \`${subtreePath}\`.  It is your responsibility to ensure these changes are merged back into \`${destRepo}\`.  Follow these handy steps!\n`;
+        markdownText += `1. \`cd ${pr.repo}\`\n`;
+        markdownText += `2. \`git checkout ${danger.github.pr.head.ref}\`\n`;
```
jtreanor: Should this be:

```suggestion
        markdownText += `2. \`git checkout -b ${danger.github.pr.head.ref}\`\n`;
```

org/github/pr-for-subtree.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // If we found changes in the subtree folder, add instructions to the PR.
+    if (containsSubtreeChanges) {
+        console.log(`PR contains changes in ${subtreePath} subtree.  Adding merge instructions as a warning.`);
+        
+        // Where the subtree changes need to go.
+        const destRepo = "WordPress-Login-Flow-Android";
+        
+        // Repo owner.
+        const org = "Automattic";
```
jtreanor: The `WordPress-Login-Flow-Android` repo is in the `wordpress-mobile` org, so we probably want this:

```suggestion
        const org = "wordpress-mobile";
```

org/github/pr-for-subtree.ts
jtreanor: Oh, my mistake actually, this is correct as is 😄

jtreanor:
> This plus two other hardcoded values, the destination repo name and subtree path to monitor, might be discoverable at runtime (well, not the subtree path, but maybe the other two 😃 ), enabling this rule to be used for any subtree.

This is a great idea and could be something to add later 😄

peril-settings.json
```diff
@@ -22,7 +22,8 @@
                 "Automattic/peril-settings@org/android.ts",
                 "Automattic/peril-settings@org/github/label.ts",
                 "Automattic/peril-settings@org/github/milestone.ts",
-                "Automattic/peril-settings@org/github/diff-size.ts"
+                "Automattic/peril-settings@org/github/diff-size.ts",
+                "Automattic/peril-settings@org/github/pr-for-subtree.ts"
```
jtreanor: Since this is currently specific to the login subtree, could we rename this to something like `check-login-subtree.ts`?

org/github/pr-for-subtree.ts
markpar: Whoops!  Good catch!  Fixed.

peril-settings.json
markpar: That's a great idea!  Done.

PR #8, opened 2019-06-17
Run tests on CircleCI

jtreanor:
This PR adds a CircleCI config to run our newly added jest tests on CircleCI (https://github.com/Automattic/peril-settings/pull/7).

It also updates `package.json` with a new dependency to support this and updates the licence to remove a warning.

### To Test

See that CircleCI has successfully run the tests on this PR.

cc @markpar

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "license": "GPLv2",
+  "license": "GPL-2.0-or-later",
   "scripts": {
     "test": "jest"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,12 +1,13 @@
 {
-  "license": "GPLv2",
+  "license": "GPL-2.0-or-later",
```
markpar: Good catch!

jtreanor: Yeah, your initial thought was correct and yarn will complain otherwise 😄

PR #14, opened 2019-08-19
Post comment with APK link on Android PRs with installable builds

jtreanor:
Adds a Peril event that will post a comment on Android PRs with an APK to test. You can see an example of this working [here](https://github.com/wordpress-mobile/WordPress-Android/pull/10404#issuecomment-522626257). Screenshot:

image

It does this by using the `status.success` webhook as follows:

- If the status check is for a CircleCI build called `Installable Build`, use the CircleCI API to check for APK artifacts.
- Post a comment (or update the existing one) to the PR with a link to download the APK.

### DIFF
--- a/org/pr/android-installable-build.ts
+++ b/org/pr/android-installable-build.ts
@@ -1,0 +1,100 @@
+import { danger } from "danger"
+import { Status } from "github-webhook-event-types"
+import fetch from "node-fetch"
+import * as url from "url";
+import * as path from "path";
+
+const CIRCLECI_TOKEN: string = process.env['CIRCLECI_TOKEN']
+const PERIL_BOT_USER_ID: number = parseInt(process.env['PERIL_BOT_USER_ID'], 10)
+
+async function circleCIArtifacts(owner: string, repo: string, buildNumber: number) {
+    const url = `https://circleci.com/api/v1.1/project/gh/${owner}/${repo}/${buildNumber}/artifacts?circle-token=${CIRCLECI_TOKEN}`
+    const res = await fetch(url)
+    if (res.ok) {
+      return res.json()
+    }
+    return []
+}
+
+export default async (status: Status) => {
+    if (status.state !== "success") {
+      return console.log(
+        `Not a successful state - got ${status.state}`
+      )
+    }
+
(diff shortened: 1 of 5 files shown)


org/pr/android-installable-build.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export default async (status: Status) => {
+    if (status.state !== "success") {
+      return console.log(
+        `Not a successful state - got ${status.state}`
+      )
+    }
+
+    if (status.context !== "ci/circleci: Installable Build") {
```
loremattei: Does this mean that we're going to call all the builds that have an apk "Installable Build"?

loremattei:
Code makes sense, though I think I miss some context about how the build actually works in order to be able to fully understand it. 
The link in your comment seems to link to this PR. 
Do we have a test PR that uses this code?

org/pr/android-installable-build.ts
jtreanor: Yes. The `context` string in the webhook is the only way for us to know what type of build this is.

Alternatively, we could check the artifacts of every CircleCI build to check if there is an APK and post a comment with that.

jtreanor:
> Code makes sense, though I think I miss some context about how the build actually works in order to be able to fully understand it.

I'll try make things a little clearer. The idea is that there is a new CircleCI job on any Android project that wants to use installable builds. It should be called `Installable Build` (open to other suggestions though) and add the `.apk` to its artifacts.

When this build completes, Peril receives a Webhook. It checks the build results and the artifacts and posts a comment to the PR to download the APK. Does that make sense?

> Do we have a test PR that uses this code?

Yes, this WPAndroid PR is using it: https://github.com/Automattic/peril-settings/pull/14#issue-308546694. I pointed out Peril instance at this branch so that it can be tested. You can see that if you re-run the CircleCI build on that PR the comment is updated.

org/pr/android-installable-build.ts
loremattei: 👍 Perfect! Thanks for clarifying. 
Makes sense to me.

loremattei:
This is awesome :-) 
:shipit:

PR #21, opened 2019-10-10
Add new action to force pending commit statuses to be successful

jtreanor:
We would like to start taking advantage of CircleCI's [approval feature](https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval) so that we can avoid running resource-intensive jobs on every push.

The problem with this is that jobs pending approval show as in progress status checks on Github, which is confusing and misleading. For example, I have added a job pending approval in https://github.com/wordpress-mobile/WordPress-iOS/pull/12654. This is what the PR looks like:

image (2)

This adds a small action to Peril to look for a list of statuses with a specific label (e.g. `ci/circleci: Installable Build/Hold`) in the pending state and use the Github API to mark is as successful.

It is difficult to test this without deploying but once the tests are passing I think its safe.

### DIFF
--- a/org/pr/force-circleci-status.ts
+++ b/org/pr/force-circleci-status.ts
@@ -1,0 +1,35 @@
+import { danger } from "danger"
+import { Status } from "github-webhook-event-types"
+
+// This is a list of the CircleCI statuses to always mark as success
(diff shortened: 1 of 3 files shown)


jtreanor:
Thank you!

PR #23, opened 2019-10-16
Add rule to check for Tracks related changes.

loremattei:
This PR adds a `check-tracks` rule that publishes a comment when changes related to `Tracks` are detected. 
The rule is theoretically general, but given the hardcoded file names, it's going to run only for WooCommerce Android.


loremattei:
Thank you @jtreanor! 
I added it :-)

PR #25, opened 2019-10-18
Run installable builds rule for WCiOS

jtreanor:
This is just a quick opt-in for WCiOS to support installable builds. Until its implemented, it will just do nothing, so no risk here.


jtreanor:
I'm just merging this to proceed with https://github.com/woocommerce/woocommerce-ios/pull/1379. There is nothing really to review here.

PR #28, opened 2020-01-03
Adds peril settings for Newspack iOS

aerych:
Adds a new setting to support installable builds in Newspack iOS.


aerych:
Thanks @jkmassel !

PR #29, opened 2020-01-22
Add temp debug stuff

loremattei:
I'm trying to figure out an issue where the `comment.json` is uploaded to CircleCI and has the right format, but it's discarded/not found by Peril, so this PR adds some debug messages to the console.

This is intended to be reverted as soon as the issue is fixed.


loremattei:
The test fails, but it seems to be unrelated to this PR. 
I just found today that the project on CircleCI had no Followers, so tests didn't run for a while now. I suppose a PR that broke the tests was merged while they weren't run. 

I'll fix the test in a different PR, just because this will make it easy to revert this one :-)

jkmassel:
I approved this because the code looks ok – not sure what's up with the tests, but don't want to block you :)

PR #34, opened 2020-01-28
Fix artifacts download on private repos

loremattei:
This PR:
- reverts the additional logging added to debug this issue (but still adds a log entry in a critical point). 
- removes a call to an API that doesn't actually seem to be in the Danger DSL. It was added here: https://github.com/Automattic/peril-settings/commit/0c7b962a56db5a478b6d9beeb498f4f2a1c830ae#diff-e03db42699566c628c96cd4785dde689, but I can't find a reference to it in the DSL: https://danger.systems/js/reference.html#DangerDSLType. Note that this fixes the tests, but only because the endpoint was not mocked. If I'm wrong and the endpoint exists, the proper fix would be to add it to the mocked object.
- adds the auth token to the API call to download artifacts. 

As usual with this repo, there's no easy way to test this. 
I tested locally with an ad-hoc JS script and it fixed the issue for the Newspack-iOS repository without breaking anything for the others :-) 
It's also compliant with the doc here: https://circleci.com/docs/api/#download-an-artifact-file

### DIFF
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -115,6 +115,10 @@ async function getDownloadCommentText(status) {
     if (res.ok) {
       const comment = await res.json()
       return comment.body
+    } else {
+      console.log(
+        `Error while trying to download comment.json: ${res.statusText}`
+      )
     }
   }
 
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -111,7 +111,8 @@ async function getDownloadCommentText(status) {
   const commentJsonArtifact = artifacts.find(artifact => artifact.path.endsWith("comment.json"))
   if (commentJsonArtifact) {
     // Download the JSON file so we can get the comment text
-    const res = await fetch(commentJsonArtifact.url)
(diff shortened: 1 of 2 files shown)


aerych:
Codewise, afaict, this looks good to me. I don't have a lot of experience with typescript so this with a grain of salt. Not confident I'm a good reviewer on this one but it sounds like the best way to test is just to commit and see and rollback if there are issues. 🤷‍♂

org/pr/installable-build.ts
```diff
@@ -117,31 +109,17 @@ async function getDownloadCommentText(status) {
...
   if (commentJsonArtifact) {
-    // <Debug: Remove me>
-    console.log(
-      `Comment JSON - '${commentJsonArtifact.url}'`
-    )
-    // </Debug: Remove me>
-    
     // Download the JSON file so we can get the comment text
-    const res = await fetch(commentJsonArtifact.url)
+    const commentJsonArtifactUrl = `${commentJsonArtifact.url}?circle-token=${CIRCLECI_TOKEN}`
```
jkmassel: After our discussion yesterday – this appears to be where all the magic happens. LGTM!

loremattei: Yep, it is :-)

PR #37, opened 2020-02-13
Skip checks for releases

loremattei:
This PR disables some Peril checks for the PRs that have the "Releases" label.
It also adds some tests for checks that were not covered. 

There's no easy way to actually test it :-(, but test cases have been added to test these changes.


loremattei:
I just added the missing part. Ready for another look @jkmassel!

PR #45, opened 2020-04-09
Update gb-mobile comment

Stojdza:


Stojdza:
Merging this. Only comment update and tests are passing.

PR #49, opened 2020-05-14
Disable milestone check for draft PRs and wip features

oguzkocer:
This PR disables the milestone check in certain conditions. If the PR is a draft, we always skip the check. This should give developers more flexibility for when they are actively working on a PR. There is now an optional label `Part of a WIP Feature` which will only disable the check if the base branch is **not** `develop` or `release/{version}`.

Disabling this check for wip features was something requested by developers and @elibud gave 👍 as long as it doesn't affect regular PRs. Since a PR needs to be merged to `develop` or a release branch before it can be shipped, I can't think of a scenario where we might miss this check with this setup. I am even questioning whether the extra label is necessary and if we should just disable the milestone checks for PRs unless they are being merged to `develop` or release branches.

If we do keep the label, I am finding it hard to find a good name for it, so if you have any suggestions to change `Part of a WIP Feature`, I'd be all for it.

I added a bunch of unit tests for these, but I still don't know what is the best way to test this without merging or using a temporary github project. (which would require setup time I'd prefer to avoid) @mokagio I think you have some experience in this area, any ideas?

_I am opening this as a draft PR because I'd like to coordinate an announcement p2 post with it. Please do review it, but don't merge it yet. Thank you!_

### DIFF
--- a/org/pr/milestone.ts
+++ b/org/pr/milestone.ts
@@ -3,6 +3,7 @@ import {warn, danger} from "danger";
 export default async () => {
     // Skip for release PRs and PRs against feature branches
     const githubLabels = danger.github.issue.labels;
+    const pr = danger.github.pr;
 
     if (githubLabels.length != 0) {
         const releases = githubLabels.some(label => label.name.includes("Releases"));
@@ -10,8 +11,10 @@ export default async () => {
             return;
         }
 
+        const targetsDevelop = pr.base.ref == "develop";
+        const targetsRelease = pr.base.ref.startsWith("release/");
         const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
-        if (wipFeature) {
+        if (!targetsDevelop && !targetsRelease && wipFeature) {
             return;
         }
     }
--- a/org/pr/milestone.ts
+++ b/org/pr/milestone.ts
@@ -1,18 +1,24 @@
 import {warn, danger} from "danger";
 
 export default async () => {
-    // Skip for release PRs and PRs against feature branches
     const githubLabels = danger.github.issue.labels;
-    const pr = danger.github.pr;
+    const currentPR = await danger.github.api.pulls.get(danger.github.thisPR);
+
+    // Skip for draft PRs
+    if (currentPR.data.draft) {
+        return;
+    }
 
     if (githubLabels.length != 0) {
+        // Skip for PRs with "Releases" label
         const releases = githubLabels.some(label => label.name.includes("Releases"));
         if (releases) {
             return;
         }
 
-        const targetsDevelop = pr.base.ref == "develop";
-        const targetsRelease = pr.base.ref.startsWith("release/");
+        // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
+        const targetsDevelop = danger.github.pr.base.ref == "develop";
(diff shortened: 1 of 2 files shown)


org/pr/milestone.ts
```diff
@@ -1,19 +1,32 @@
...
+
+        // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
+        const targetsDevelop = danger.github.pr.base.ref == "develop";
+        const targetsRelease = danger.github.pr.base.ref.startsWith("release/");
+        const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
+        if (!targetsDevelop && !targetsRelease && wipFeature) {
+            return;
+        }
     }
-
```
mokagio, line 13: Yes! 🤓

org/pr/milestone.ts
```diff
@@ -1,19 +1,32 @@
 import {warn, danger} from "danger";
 
 export default async () => {
-    // Skip for release PRs
     const githubLabels = danger.github.issue.labels;
+    const currentPR = await danger.github.api.pulls.get(danger.github.thisPR);
```
loremattei, line 5: Does this work? 
Looking at the DSL, it seems that there's no `get` method that accepts only the PR number as parameter. It may still work if the other parameters are provided by the context/wrapper, so I just want to double check that it has been verified.

oguzkocer: @loremattei Unfortunately I don't know if this works or not. However, `danger.github.thisPR` has the following comment in [GitHubDSL's documentation](https://danger.systems/js/reference.html#GitHubDSL):

```
GitHubDSL {
  ...
  // The PR metadata specifically formatted for using with the GitHub API client
  thisPR: GitHubAPIPR
  ...
}
```

And here is the [documentation for GitHubAPIPR](https://danger.systems/js/reference.html#GitHubAPIPR) which it links to:

```
// Provides the current PR in an easily used way for params in `github.api` calls
GitHubAPIPR {
  // The PR number
  number: number
  // The repo owner
  owner: string
  // The repo name
  repo: string
}
```

Am I maybe missing something?

oguzkocer:
@mokagio I have been unable to run the command you provided, is there maybe a documentation I can look at for it? I tried to find one, but nothing similar is showing up in my searches.

Could you also test the PR with those instructions?

oguzkocer:
@mokagio @loremattei Could you follow up on this PR this week when you get a chance?

org/pr/milestone.ts
loremattei: Ah, cool. This is perfect and matches the function signature: https://octokit.github.io/rest.js/v17#pulls-get. 👍

org/pr/milestone.ts
```diff
@@ -1,19 +1,32 @@
...
+
+    // Skip for draft PRs
+    if (currentPR.data.draft) {
+        return;
+    }
```
mokagio, line 10: Here's something I discovered while testing this against real PR.

If a PR was a draft and then it's closed, `currentPR.data.draft` won't be `null`/`false` and we'll hit the `return` here.

Obviously not an issue, because we don't care about running these checks on closed PRs, but I still found interesting. My guess would be that GitHub doesn't remove that state information when closing the PR, which makes sense.

oguzkocer:
Thank you both for the reviews!

PR #56, opened 2020-10-27
Remove Gutenberg exception when checking pod references on iOS

loremattei:
This PR removes the exception for Gutenberg when checking that the pods are referenced by version. 

To Test:
- Check that CI is green.

### DIFF
--- a/org/pr/ios-macos.ts
+++ b/org/pr/ios-macos.ts
@@ -18,10 +18,7 @@ export default async () => {
     const podfileContents = await danger.github.utils.fileContents("Podfile");
     const matches = podfileContents.match(/^[^#]*:commit/gm);
     if (matches !== null) {
-        const nonGutenbergMatches = matches.filter(m => !m.includes("gutenberg"));
-        if (nonGutenbergMatches.length > 0) {
-            fail("Podfile: reference to a commit hash");
-        }
+        fail("Podfile: reference to a commit hash");
     }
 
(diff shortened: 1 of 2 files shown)


org/pr/ios-macos.ts
```diff
@@ -18,10 +18,7 @@ export default async () => {
     const podfileContents = await danger.github.utils.fileContents("Podfile");
     const matches = podfileContents.match(/^[^#]*:commit/gm);
```
ceyhun, line 19: I think this might give a false negative for gutenberg because of this line:
https://github.com/wordpress-mobile/WordPress-iOS/blob/342f9811440d740767136a42f0beac624051f309/Podfile#L133

I'm not really knowledgable with Ruby so I'm not sure how to keep the functionality same but eliminate the false negative by modifying the `Podfile`. Maybe we can modify the regex in a way that doesn't catch this line but other cases.

ceyhun (changes requested):

org/pr/ios-macos.ts
loremattei: Good catch @ceyhun! Thank you!
I've updated the regex so that it won't catch `[:commit]` and I also added a few tests to make sure it worked.

ceyhun (approved):
Looks great! Thank you!

loremattei:
Thanks for the review @ceyhun!

PR #57, opened 2020-11-11
Add checks for changes to the release notes in a release branch

loremattei:
I've noticed that sometimes changes to RELEASE-NOTES.txt are merged into the release branch thinking they will be reflected in the final release notes. Due to how our current flow for translations works, this is not the case and a manual update is required in order to try to update the release notes after the code freeze. 

This PR introduces a new check that prints a warning on PRs that:
-  target the release branch, and 
- contain changes to RELEASE-NOTES.txt. 

To Test:
Testing changes to Peril rules is not easy. Because of this I've added some tests to cover the most common cases. 

 - Make sure CI is green. 
 - Verify that the mocked API responses match the actual GitHub API.

### DIFF
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -21,14 +21,13 @@ export default async () => {
     ];
 
     // Skip if not targeting a release branch
-    if (checkedBranches.filter((branch) => pr.base.ref.startsWith(branch)).length == 0) {
+    if (!checkedBranches.some(branch => pr.base.ref.startsWith(branch))) {
         return;
     }
 
     // Skip for PR labeled as "Releases"
     if (githubLabels.length != 0) {
-        const releases = githubLabels.some(label => label.name.includes("Releases"));
-        if (releases) {
+        if (githubLabels.some(label => label.name.includes("Releases"))) {
             return;
         }
     }
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -3,6 +3,7 @@
 // This file should typically not be modified after code freeze (i.e. on the release branch).
 
 import {warn, danger} from "danger";
+import { Stream } from "stream";
 
 export default async () => {
     const githubLabels = danger.github.issue.labels;
@@ -14,8 +15,13 @@ export default async () => {
         "RELEASE-NOTES.txt"
     ];
 
-    // Skip if not targeting a release branch 
-    if (!pr.base.ref.startsWith("release/")) {
+    const checkedBranches: string[] = [
+        "release/",
+        "hotfix/"
+    ];
+
+    // Skip if not targeting a release branch
+    if (checkedBranches.filter((branch) => pr.base.ref.startsWith(branch)).length == 0) {
         return;
     }
 
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -10,7 +10,7 @@ export default async () => {
         "RELEASE-NOTES.txt"
     ];
 
-    // Skip if not on a release branch 
+    // Skip if not targeting a release branch 
     if (!pr.base.ref.startsWith("release/")) {
         return;
     }
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -1,3 +1,7 @@
+// This Peril rule will warn if the `RELEASE-NOTES.txt` file has been changed in a PR targeting a `release/*` branch.
+//
+// This file should typically not be modified after code freeze (i.e. on the release branch).
+
 import {warn, danger} from "danger";
 
 export default async () => {
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -27,7 +27,7 @@ export default async () => {
         if (modifiedFiles.some(f => f.includes(releaseNotes))) {
             let messageText: string;
 
-            messageText = "This PR contains changes to RELEASE_NOTES.txt.\n";
+            messageText = "This PR contains changes to \`RELEASE_NOTES.txt\`.\n";
             messageText += "Note that these changes won't affect the final version of the release notes as this version is in code freeze.\n";
             messageText += "Please, get in touch with a release manager if you want to update the final release notes.";
 
--- a/org/pr/release-notes.ts
+++ b/org/pr/release-notes.ts
@@ -1,0 +1,37 @@
+import {warn, danger} from "danger";
+
+export default async () => {
+    const githubLabels = danger.github.issue.labels;
+    const modifiedFiles = danger.git.modified_files;
+    const pr = danger.github.pr;
+
+    // List of release notes files
+    const releaseNotesFiles: string[] = [
+        "RELEASE-NOTES.txt"
+    ];
+
+    // Skip if not on a release branch 
+    if (!pr.base.ref.startsWith("release/")) {
+        return;
+    }
+
+    // Skip for PR labeled as "Releases"
+    if (githubLabels.length != 0) {
+        const releases = githubLabels.some(label => label.name.includes("Releases"));
+        if (releases) {
+            return;
+        }
+    }
+    
+    for (let releaseNotes of releaseNotesFiles) {
+        if (modifiedFiles.some(f => f.includes(releaseNotes))) {
+            let messageText: string;
+
+            messageText = "This PR contains changes to RELEASE_NOTES.txt.\n";
+            messageText += "Note that these changes won't affect the final version of the release notes as this version is in code freeze.\n";
+            messageText += "Please, get in touch with a release manager if you want to update the final release notes.";
+
+            warn(messageText);
+        }
+    }
+};
--- a/peril-settings.json
+++ b/peril-settings.json
@@ -78,7 +78,8 @@
                 "Automattic/peril-settings@org/pr/ios-macos.ts",
                 "Automattic/peril-settings@org/pr/label.ts",
                 "Automattic/peril-settings@org/pr/milestone.ts",
-                "Automattic/peril-settings@org/pr/diff-size.ts"
+                "Automattic/peril-settings@org/pr/diff-size.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
             ],
             "status": [
                 "Automattic/peril-settings@org/pr/installable-build.ts"
@@ -87,13 +88,15 @@
         "Automattic/simplenote-macos": {
             "pull_request, pull_request.labeled, pull_request.unlabeled": [
                 "Automattic/peril-settings@org/pr/ios-macos.ts",
-                "Automattic/peril-settings@org/pr/label.ts"
+                "Automattic/peril-settings@org/pr/label.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
             ]
         },
         "Automattic/simplenote-android": {
             "pull_request, pull_request.labeled, pull_request.unlabeled": [
                 "Automattic/peril-settings@org/pr/android.ts",
-                "Automattic/peril-settings@org/pr/label.ts"
+                "Automattic/peril-settings@org/pr/label.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
             ],
             "status": [
                 "Automattic/peril-settings@org/pr/installable-build.ts"
--- a/peril-settings.json
+++ b/peril-settings.json
@@ -12,7 +12,8 @@
                 "Automattic/peril-settings@org/pr/ios-macos.ts",
                 "Automattic/peril-settings@org/pr/label.ts",
                 "Automattic/peril-settings@org/pr/milestone.ts",
-                "Automattic/peril-settings@org/pr/diff-size.ts"
+                "Automattic/peril-settings@org/pr/diff-size.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
             ],
             "issues.opened, issues.labeled, issues.unlabeled": [
                 "Automattic/peril-settings@org/issue/label.ts"
@@ -28,7 +29,8 @@
                 "Automattic/peril-settings@org/pr/label.ts",
                 "Automattic/peril-settings@org/pr/milestone.ts",
                 "Automattic/peril-settings@org/pr/diff-size.ts",
-                "Automattic/peril-settings@org/pr/check-subtrees.ts"
+                "Automattic/peril-settings@org/pr/check-subtrees.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
             ],
             "issues.opened, issues.labeled, issues.unlabeled": [
                 "Automattic/peril-settings@org/issue/label.ts"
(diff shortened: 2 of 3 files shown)


org/pr/release-notes.ts
```diff
@@ -0,0 +1,37 @@
...
+    const githubLabels = danger.github.issue.labels;
+    const modifiedFiles = danger.git.modified_files;
+    const pr = danger.github.pr;
+
+    // List of release notes files
+    const releaseNotesFiles: string[] = [
+        "RELEASE-NOTES.txt"
+    ];
+
+    // Skip if not on a release branch
```
AliSoftware, line 13:
```suggestion
    // Skip if not targeting a release branch 
```
Just to be clear that we only analyse PRs targeting the `release/*` branches, so PRs done after code freeze, but we won't check targets _from_ `release/*` (e.g. to `develop` or to any other branch)

org/pr/release-notes.ts
```diff
@@ -0,0 +1,37 @@
...
+        return;
+    }
+
+    // Skip for PR labeled as "Releases"
+    if (githubLabels.length != 0) {
+        const releases = githubLabels.some(label => label.name.includes("Releases"));
+        if (releases) {
+            return;
+        }
+    }
```
AliSoftware, line 24: Mmmh maybe I'm not completely up-to-date with when we use this Releases label and when not (I thought it was equally applied both for PRs targetting the release branch – like fixes made during the code freeze – and for PRs merging the release branch (back into develop after new betas, back into develop + trunk after store submission)?

Or should this label only be used for the PRs from release to develop and trunk that are happening after store submission, but should not be used for PRs targetting the current code-frozen release?

I fear that people might easily add that "Releases" label to PRs fixing code-freeze-bugs and thus targetting the `release/*` branch anyway, which would make this test to be skipped if so; not sure why we would want to test the label anyway if we already test that the target branch is `release/*` 🤔

org/pr/release-notes.ts
```diff
@@ -0,0 +1,37 @@
...
+    
+    for (let releaseNotes of releaseNotesFiles) {
+        if (modifiedFiles.some(f => f.includes(releaseNotes))) {
+            let messageText: string;
+
+            messageText = "This PR contains changes to RELEASE_NOTES.txt.\n";
```
AliSoftware, line 30: nit
```suggestion
            messageText = "This PR contains changes to \`RELEASE_NOTES.txt\`.\n";
```

peril-settings.json
```diff
@@ -63,7 +66,8 @@
                 "Automattic/peril-settings@org/pr/milestone.ts",
                 "Automattic/peril-settings@org/pr/diff-size.ts",
                 "Automattic/peril-settings@org/pr/check-subtrees.ts",
-                "Automattic/peril-settings@org/pr/check-tracks.ts"
+                "Automattic/peril-settings@org/pr/check-tracks.ts",
+                "Automattic/peril-settings@org/pr/release-notes.ts"
```
AliSoftware, line 70: We don't have `RELEASE_NOTES.txt` for `simplenote-ios`, `simplenote-macos` and `simplenote-android`?

tests/release-notes-test.ts
```diff
@@ -0,0 +1,68 @@
...
+            },
+        },
+    };
+});
+
+describe("release notes checks", () => {
+    it("warns when RELEASE-NOTES.txt is changed in a PR which targets a release branch", async () => {
+        await checkReleaseNotes();
+        
+        expect(dm.warn).toHaveBeenCalledWith(expect.stringContaining("This PR contains changes to RELEASE_NOTES.txt.\n"));
```
AliSoftware, line 34: nit to only apply if you applied the previous one too
```suggestion
        expect(dm.warn).toHaveBeenCalledWith(expect.stringContaining("This PR contains changes to \`RELEASE_NOTES.txt\`.\n"));
```

tests/release-notes-test.ts
```diff
@@ -0,0 +1,68 @@
...
+                "file1.swift",
+                "file2.php",
+                "file3.kt"
+        ]
+
+        await checkReleaseNotes();
+        
+        expect(dm.warn).not.toHaveBeenCalled();
+    })
+})
```
AliSoftware, line 68: nit: CRLF at EOF (to avoid potential oddities with git and diff and other tooling)
```suggestion
})

```

AliSoftware (commented):
Mostly LGTM, but just not sure we really want to skip the test depending on the "Releases" GH label though?

AliSoftware:
PS: Super-nit but wondering if the name of the rule is the best one, given that maybe in the future we'd want a separate rule to do something else with the release notes (like lint the content of the file for typos or grammar check it with an English linter or whatever); what about `release-notes-code-freeze.ts` for example?

_(this is my OCD speaking so feel free to ignore)_

tests/release-notes-test.ts
mokagio: I once did some research into the rationale for this and came up with the following pre-def comment that I sometimes add to code reviews:

The convention of having a new line at the end of the files comes from an old C decision that has been passed down through Unix history:

> A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character.
> 
> Since this is a “shall” clause, we must emit a diagnostic message for a violation of this rule.

So, it turns out that, according to POSIX, every text file (including Ruby and JavaScript source files) should end with a `\n`, or “newline” (not “a new line”) character. This acts as the `eol`, or the “end of line” character. It is a line “terminator”. ([Source](https://thoughtbot.com/blog/no-newline-at-end-of-file))

If anything, addressing these also spares us from that annoying red indicator when reading the file on GitHub. I keep talking about `.editorconfig` and how it can help with this, but never do it 😞

org/pr/release-notes.ts
mokagio: My understanding of the "Releases" label is that we use it to mark the PRs that merge beta and finalized versions back into `develop` and `trunk`.

That helps with another warning, the one for the code generated [`Localizable.strings`](https://github.com/Automattic/peril-settings/blob/685783686a07f5a09ee8d65acbf96335f7b0b146/org/pr/ios-macos.ts#L32-L45).

mokagio: > I fear that people might easily add that "Releases" label to PRs fixing code-freeze-bug

Working with the Simplenote team, I've never seen this pattern. 

That doesn't mean that it wouldn't be possible, though. If we're concerned about it, maybe the label could have a more explicit name, like "Release Merge".

peril-settings.json
mokagio: We do have them in the Simplenote repos, too.

org/pr/release-notes.ts
```diff
@@ -0,0 +1,37 @@
...
+            messageText += "Note that these changes won't affect the final version of the release notes as this version is in code freeze.\n";
+            messageText += "Please, get in touch with a release manager if you want to update the final release notes.";
```
mokagio, line 32: I like this wording. Hopefully it helps people out.

mokagio (commented):
Other than for the fact that the Simplenote repos could use these checks, too , this LGTM 👍

org/pr/release-notes.ts
AliSoftware: TBH I think I'm guilty of having used that `Releases` label that way (i.e. used it also for PRs merging fixes into the code-frozen `release/*` branch), not really knowing what its real definition/purpose was. Not sure if "Release Merge" would help disambiguate.

This could actually be another Peril check: ensure that if the Releases label is set, the PR is from `release/*` and targeting either `develop` or `trunk`, but not a PR targeting `release/*` itself.
Bonus question: are there PRs from `release/*` to `develop` or `trunk` that we _don't_ label with the `Releases` label? (maybe the initial one immediately after code freeze?) if we label all `release/* -> (develop|trunk)` with the label maybe we can automate that label (with Peril? can it _set_ labels?)

mokagio: > This could actually be another Peril check: ensure that if the Releases label is set, the PR is from `release/*` and targeting either `develop` or `trunk`, but not a PR targeting `release/*` itself.

Great idea!

mokagio: > are there PRs from `release/*` to `develop` or `trunk` that we _don't_ label with the Releases label? (maybe the initial one immediately after code freeze?) if we label all `release/* -> (develop|trunk)` with the label maybe we can automate that label

I can't think of any occasion in which I didn't have to use Release. Every merge of the release branch in `develop` corresponds to a new build that's being distributed, as far as I know.

> (with Peril? can it set labels?)

It can. Peril has access to an authenticated GitHub API wrapper via Danger.

loremattei:
> Super-nit but wondering if the name of the rule is the best one, given that maybe in the future we'd want a separate rule to do something else with the release notes (like lint the content of the file for typos or grammar check it with an English linter or whatever); what about release-notes-code-freeze.ts for example?

In many files, we have multiple rules per file, grouped by theme. So we may add new rules related to `Release Notes` to the same file. As usual, there are pros and cons: files becoming larger vs many different files in the repository. I think that it's up to us to find the right balance. 
Since the rule name visibility is restricted to this repo, I think we can keep this name now and then iterate on it if we'll need to add many rules related to the `release notes`. 
Does it make sense to you?

loremattei:
@AliSoftware @mokagio thanks for the review. I think I've addressed all the points, so this should be ready for another pass.

AliSoftware:
> Does it make sense to you?

Yep, let's keep the current name then 👍

org/pr/release-notes.ts
```diff
@@ -0,0 +1,37 @@
+import {warn, danger} from "danger";
+
+export default async () => {
```
AliSoftware, line 3: Maybe include a quick general description about what this Peril rule is about at the top of the file?
```suggestion

// This Peril rule will warn if the `RELEASE-NOTES.txt` file has been changed in a pr targeting a `release/*` branch.
//
// This file should typically not be modified after code freeze (i.e. on the release branch) because the release notes
// have already been sent to Editorial then up for translation at this point so due to our process about this,
// any changes to the file won't be automatically included in the corresponding release in practice.

export default async () => {
```

tests/release-notes-test.ts
```diff
@@ -0,0 +1,69 @@
...
+        git: {
+            modified_files: ["RELEASE-NOTES.txt"],
+            created_files: [],
+            deleted_files: []
+        },
+        github: {
+            pr: {
+                base: {
+                    ref: "release/1.0",
+                }
```
AliSoftware, line 21: I think it could be useful to consider more test cases for some other PR branch configurations, including:

 - `head: { ref: "release/1.0" }, base: { ref: "develop" }` : a PR from release to develop, which should be _allowed_ to have RELEASE-NOTES.txt changes (especially after we update them once Editorial get back to us
 - `base: { ref: "release/merge-1.0-to-develop" }` : type of PRs we use for conflict resolution. Do we want the rule to forbid `RELEASE-NOTES.txt` changes for those? Though not sure what most people use as naming convention for those, we might not be all consistent, but it's worth including a Unit Test for imho to at least not forget that use case and test our expected behavior against it
 - `base: { ref: "hotfix/1.0.1" }`: shouldn't we enable the check (and forbid release notes changes) for PRs targeting a hotfix like we do for a release?

AliSoftware (commented):
Open questions:

 - What about PRs targetting hotfixes?
 - And how should we treat conflict-resolution intermediate branches we sometimes use?

org/pr/release-notes.ts
loremattei: Right! I pushed a shorter description which outlines what the rule does, but doesn't go into the process. I think that adding notes about process becomes "documentation about the process" that then needs to be maintained 😅
I prefer to decouple stuff: we have a rule that does something, and we integrate it where we need it.

tests/release-notes-test.ts
loremattei: > head: { ref: "release/1.0" }, base: { ref: "develop" } : a PR from release to develop, which should be allowed to have RELEASE-NOTES.txt changes (especially after we update them once Editorial get back to us)

There's already a test with `develop` as the base branch and the rule doesn't mind the `head` branch at all, so isn't it enough? 🤔

> base: { ref: "release/merge-1.0-to-develop" } : type of PRs we use for conflict resolution. Do we want the rule to forbid RELEASE-NOTES.txt changes for those?

The rule doesn't mind the `head` branch, so I'm not sure this would be useful. 

> shouldn't we enable the check (and forbid release notes changes) for PRs targeting a hotfix like we do for a release?

Yup! I believe nobody expects the release-notes to be translated for a hotfix... but I agree that adding it doesn't hurt. I've updated the PR.

AliSoftware: > the rule doesn't mind the head branch at all, so isn't it enough? 🤔

Well strictly speaking, saying that means you rely on the current implementation details to write your test (as opposed to thinking TDD 😅 ). Which means that if for whatever reason we refactor the rule a bit in the future, thinking we'd improve its performance without changing the behavior, we could accidentally check the wrong branch after our refactor, or use e.g. an hypothetical new method like `github.pr.branches.contains('release')` new property, instead of specifically `head` or `base`, and break the behavior without the test catching it 😉

loremattei: > Well strictly speaking, saying that means you rely on the current implementation details to write your test 

Well, tbh, I think it's the opposite: I'm not looking at the implementation details at all ;-) Let's step back a bit: this rule is about checking that a certain type of changes  don't get merged into a release branch. From where they come, it doesn't matter at all (if you don't think about the implementation), because there are no special cases where we want this behaviour to change depending on the head branch.
But I agree that if you look at the implementation details, you may be worried that a branch which has `release` in the name can become a problem in the future if we make a mistake with the code :-)

loremattei:
Oh, good point! Code updated.

loremattei:
Hey @AliSoftware! Do you think this is ready?

org/pr/release-notes.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Skip for PR labeled as "Releases"
+    if (githubLabels.length != 0) {
+        if (githubLabels.some(label => label.name.includes("Releases"))) {
+            return;
+        }
+    }
+    
+    for (let releaseNotes of releaseNotesFiles) {
+        if (modifiedFiles.some(f => f.includes(releaseNotes))) {
+            let messageText: string;
```
AliSoftware, line 37: Being used to Swift, Ruby and similar languages, this feels so strange to see a type like `string` being lowercased and not titlecased here 😅

org/pr/release-notes.ts
AliSoftware: I like the copy too.
Just a small nit, now that this warning is also used for hotfixes, do you think it might be confusing in those cases to see the warning speak about "this version is in code freeze"? (not a big deal, I think people will get the intent even for hotfixes anyway, but just thought I'd raise the question)

AliSoftware (approved):
LGTM!

Just an open question about the warning copy now that it will also appear in the context of hotfixes, but not really a blocker and I think people will get it anyway, so I'm approving the PR already 👍

org/pr/release-notes.ts
loremattei: 😄

PR #58, opened 2020-11-18
Add rule to check that PRs in Android libraries update the version

mokagio:
Adds a rule to make sure that every PR towards `develop` or `trunk` in Android libraries bumps the library version.

You can test this using the instructions in https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/49.

### DIFF
--- a/org/pr/android-library.ts
+++ b/org/pr/android-library.ts
@@ -1,0 +1,30 @@
+import {fail, danger} from "danger";
+
+// We want each PR towards develop or trunk of our Android libraries to bump
+// the library version. This check makes sure of it.
+export default async () => {
+  const baseBranch = danger.github.pr.base.ref
+  if (baseBranch != "develop" && baseBranch != "trunk") { return }
+
+  // Android projects have more than one build.gradle file, to be able to run
+  // this check on any library, lets's just look into _all_ the build.gradle
+  // files.
+  const modified = danger.git.modified_files
+    .filter((path: string) => path.endsWith('build.gradle'))
+
+  for (let file of modified) {
+    const diff = await danger.git.diffForFile(file)
+
+    // We could be more refined and actually parse the .before and .after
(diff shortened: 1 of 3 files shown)


org/pr/android-library.ts
```diff
@@ -0,0 +1,30 @@
...
+  const baseBranch = danger.github.pr.base.ref
+  if (baseBranch != "develop" && baseBranch != "trunk") { return }
+
+  // Android projects have more than one build.gradle file, to be able to run
+  // this check on any library, lets's just look into _all_ the build.gradle
+  // files.
+  const modified = danger.git.modified_files
+    .filter((path: string) => path.endsWith('build.gradle'))
+
+  for (let file of modified) {
```
mokagio, line 15: I would have preferred not to use a for loop, I'm more into collection operations, but I didn't know how to write one because of the `await` inside.

Anyways, this does the job and it's quite tidy, too.

org/pr/android-library.ts
```diff
@@ -0,0 +1,30 @@
...
+    const diff = await danger.git.diffForFile(file)
+
+    // We could be more refiened and actually pares the .before and .after
```
loremattei, line 18:
```suggestion
    // We could be more refined and actually parse the .before and .after
```

org/pr/android-library.ts
```diff
@@ -0,0 +1,30 @@
...
+    // contents to extract the versionName value and do a SemVer check on it to
+    // make sure it's appropriate.
+    //
```
loremattei, line 21: I agree. I'm fine with deploying this first solution now and then iterate on it in the future.

loremattei (approved):
I left a nitpicky suggestion and I agree that we can iterate on it to make it smarter in the future, but it looks good to me! 
:shipit:

PR #62, opened 2021-01-08
Add fluxc/Optional Tests

oguzkocer:
This PR adds `fluxc/Optional Tests` to hold contexts so we can mark them as green. It also improves the unit tests for this so multiple contexts can be tested.

I suggest utilizing the `Hide whitespace changes` option for the diff as most of the changes are for indentation level change.

Screen Shot 2021-01-08 at 4 31 20 PM


jkmassel (approved):
LGTM

PR #63, opened 2021-01-19
Adds WordPress-FluxC-Android repo to peril settings

oguzkocer:
Since we are currently only interested in marking optional tests as green as implemented in #62, I thought only adding the `status` property would be enough. Admittedly I don't know anything about this json, so I just made an educated guess using the information about the other repos.


jkmassel (approved):
This looks correct to me – let's try it!

:shipit:

PR #64, opened 2021-01-28
Feature/add test for android strings

loremattei:
Over the last few Android releases, we have had a recurring issue where entries which reference other string resources were added to `strings.xml` without the `translatable=false` attribute. As a result, this entries were uploaded to GlotPress and they caused some confusion with translators and some bugs in some languages. 

This PR adds a Peril check that raises a warning when the above scenario happens. 

**Note**
Working on this, I noticed that we have checks for Android release notes in `Android.ts` and `release-notes.ts`. I'll open a follow up PR to reorganize the checks across the proper files. I'm not doing in in this PR to make the review easier.


tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    return diffString;
+}
+
+beforeEach(() => {
+    dm.warn = jest.fn().mockReturnValue(true);
+});
+
+describe("string checks", () => {
+    it("Warn when an entry which references another string resources is added without the translatable=false attribute", async () => {
```
AliSoftware, line 24: nit: the `it` syntax used in most test frameworks like here are named that way because they are meant to be read as English sentences 😉 

```suggestion
    it("warns when an entry which references another string resources is added without the translatable=false attribute", async () => {
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+        
+        // Run the checks
+        await androidChecks();
+        
+        // Check that the no message is shown.
+        expect(dm.warn).not.toHaveBeenCalledWith();
+    })
+
+    it("Don't warn when strings resources are added to strings.xml", async () => {
```
AliSoftware, line 60:
```suggestion
    it("does not warn when strings resources are added to strings.xml", async () => {
```

AliSoftware, line 60:
```suggestion
    it("does not warn when strings resources are added to strings.xml", async () => {
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+        
+        // Run the checks
+        await androidChecks();
+        
+        // Check that the no message is shown.
+        expect(dm.warn).not.toHaveBeenCalledWith();
+    })
+
+    it("Don't warn when there are no changes in strings.xml", async () => {
```
AliSoftware, line 78:
```suggestion
    it("does not warn when there are no changes in strings.xml", async () => {
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+        
+        // Run the checks
+        await androidChecks();
+        
+        // Check that the no message is shown.
+        expect(dm.warn).not.toHaveBeenCalledWith();
+    })
+
+    it("Warn when there are changes in metadata/release-notes.txt but not in metadata/PlayStoreStrings.po", async () => {
```
AliSoftware, line 95:
```suggestion
    it("warns when there are changes in metadata/release-notes.txt but not in metadata/PlayStoreStrings.po", async () => {
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+        
+        // Run the checks
+        await androidChecks();
+        
+        // Check that the instructions appear correct.
+        expect(dm.warn).toHaveBeenCalledWith(expect.stringContaining("The PlayStoreStrings.po file must be updated any time changes are made to release notes"));
+    })
+
+    it("Don't warn when there are changes in metadata/release-notes.txt and in metadata/PlayStoreStrings.po", async () => {
```
AliSoftware, line 112:
```suggestion
    it("does not warn when there are changes in both metadata/release-notes.txt and in metadata/PlayStoreStrings.po", async () => {
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        expect(dm.warn).not.toHaveBeenCalledWith();
+    })
+
+    it("Don't warn when there are no changes in strings.xml", async () => {
+        // Create mocks
+        dm.danger = {
+            github: { pr: { base: { ref: "a-branch" } } },
+            git: {
+                modified_files: ["a-file.kt"],
+                diffForFile: async () => { return { added: '' } }
```
AliSoftware, line 84: Maybe provide a dummy diff for the mock here, so that it at least simulates an actual change in the `.kt` file?

Otherwise that test could go green if our implementation were to erroneously test all the files (and not just `strings.xml`) but skip files with empty diffs, seeing them as unchanged, while the real reason why we want it to be green is because there's no `strings.xml`, not that there are not changes in any files including kt files.

```suggestion
                diffForFile: async () => { return { added: '+++ getString(R.string.example),' } }
```

tests/android-test.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+        
+        // Run the checks
+        await androidChecks();
+        
+        // Check that the instructions appear correct.
+        expect(dm.warn).not.toHaveBeenCalledWith();
+    })
+
+    it("Don't warn when there are changes in /release-notes.txt", async () => {
```
AliSoftware, line 129:
```suggestion
    it("does not warn when there are changes in /release-notes.txt", async () => {
```

AliSoftware (approved):
Left a couple of nits:
 - Maybe add a dummy diff to the .kt file to ensure it's a true green test pass and not a false positive.
 - The rest are just nits about test case naming conventions, nbd.

Otherwise, LGTM, so approving anyway to not block the merge.

tests/android-test.ts
loremattei: TIL! Thank you!

loremattei:
Thank you @AliSoftware! 🙇‍♂️ 
All your feedback should be addressed now!

AliSoftware:
:shipit:

PR #65, opened 2021-01-28
Add Tracks checks to WooCommerce iOS

loremattei:
This PR extends the Tracks check we have on WooCommerce Android to WooCommerce iOS. 
On WooCommerce iOS we only want to monitor changes to `WooAnalyticsStat.swift`, so adding it to the current logic should to be enough.


AliSoftware (approved):
LGTM!

PR #66, opened 2021-02-01
Fix Android translatable strings check

loremattei:
This PR fixes two bugs with the Android translatable strings check: 
- It fixes string interpolation in the warning message.
- It fixes the string pattern that the check uses (and the related tests). 

**To Test:**
Changes to Peril rules can't be easily tested, so: 
- Check that the code changes look good. 
- Check that CI is green.

### DIFF
--- a/org/pr/android.ts
+++ b/org/pr/android.ts
@@ -18,7 +18,7 @@ export default async () => {
         const stringDiffs = await danger.git.diffForFile(file)
 
         for (let stringDiff of stringDiffs.added.replace(/\r/g, "").split(/\n/)) {
-            if (stringDiff.includes("@string/") && (!stringDiff.includes("translatable=false"))) {
+            if (stringDiff.includes("@string/") && (!stringDiff.includes("translatable=\"false\""))) {
                 let markdownText: string;
 
                 markdownText = "This PR adds a translatable entry to \`strings.xml\` which references another string resource: this usually causes issues with translations."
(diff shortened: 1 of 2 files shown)


org/pr/android.ts
```diff
@@ -18,12 +18,12 @@ export default async () => {
...
         for (let stringDiff of stringDiffs.added.replace(/\r/g, "").split(/\n/)) {
-            if (stringDiff.includes("@string/") && (!stringDiff.includes("translatable=false"))) {
+            if (stringDiff.includes("@string/") && (!stringDiff.includes("translatable=\"false\""))) {
                 let markdownText: string;
 
                 markdownText = "This PR adds a translatable entry to \`strings.xml\` which references another string resource: this usually causes issues with translations."
-                markdownText += "Please make sure to set the \`translatable=false\` attribute here:"
-                markdownText += "\`` + stringDiff + `\`"
+                markdownText += "Please make sure to set the \`translatable=\"false\"\` attribute here:"
+                markdownText += `\`${stringDiff}\``
```
AliSoftware, line 26: TIL that backticks are used for string interpolation in TypeScript 👍

tests/android-test.ts
```diff
@@ -54,7 +58,7 @@ describe("string checks", () => {
         await androidChecks();
         
         // Check that the no message is shown.
-        expect(dm.warn).not.toHaveBeenCalledWith();
+        expect(dm.warn).not.toHaveBeenCalled();
```
AliSoftware, line 61: Ah! That's what made the test refuse to go red! Good catch! ☕

AliSoftware (approved):

PR #67, opened 2021-02-16
Add warning on soon-closing milestones

loremattei:
This PR adds a new check on milestones which warns when a milestone is assigned _or_ the PR is modified at a time close to the milestone's due date. 
The initial idea was to post the warning only when a milestone is assigned, but there's no way to get the previous state on Peril, so we can't know the action which triggered Peril.
Anyway, I think that having the warning published also on PRs that are worked on the last days of the cycle has a value. 

There's no easy way to test these changes, so I added some tests based on the [GitHub API documentation](https://docs.github.com/en/rest/reference/issues).

### DIFF
--- a/org/pr/milestone.ts
+++ b/org/pr/milestone.ts
@@ -30,5 +30,24 @@ export default async () => {
     // Warn if the PR doesn't have a milestone
     if (currentPR.data.milestone == null) {
         warn("PR is not assigned to a milestone.");
+        return;
+    }
+
+    // Warn if the milestone is closing in less than 4 days
+    const warningDays = 4;
+    const warningThreshold = warningDays * 1000 * 3600 * 24; // Convert days to milliseconds
+    if (currentPR.data.milestone.due_on != null) {
+        const today = new Date();
+        let dueDate : Date;
+        dueDate = new Date();
+        dueDate.setTime(Date.parse(currentPR.data.milestone.due_on));
+        
+        if ((dueDate.getTime() - today.getTime()) <= warningThreshold) {
+            let messageText : string;
+
+            messageText = `This PR is assigned to a milestone which is closing in less than ${warningDays} days\n`;
+            messageText += `Please, make sure to get it merged by then or assign it to a later expiring milestone`;
+            warn(messageText)
+        }
     }
 };
--- a/tests/pr-milestone.test.ts
+++ b/tests/pr-milestone.test.ts
@@ -117,4 +117,32 @@ describe("PR milestone checks", () => {
 
         expect(dm.warn).not.toHaveBeenCalled();
     })
+
+    for(let i = 0; i <= 4; i++) {
+        it("does warn when the milestone is present and closing in 4 days or less", async () => {
+            var closeDate = new Date();
+            closeDate.setDate(closeDate.getDate() + i);
+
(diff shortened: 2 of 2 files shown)


jkmassel (approved):

org/pr/milestone.ts
```diff
@@ -30,5 +30,24 @@ export default async () => {
...
+    }
+
+    // Warn if the milestone is closing in less than 4 days
+    const warningDays = 4;
+    const warningThreshold = warningDays * 1000 * 3600 * 24; // Convert days to milliseconds
+    if (currentPR.data.milestone.due_on != null) {
+        const today = new Date();
+        let dueDate : Date;
+        dueDate = new Date();
+        dueDate.setTime(Date.parse(currentPR.data.milestone.due_on));
```
AliSoftware, line 43: Open question since I don't know much TypeScript so maybe it works as expected, but just wanted to raise the question: how does this handle DST and timezones? As I understand it, dueDate is in an ISO string in the API but `today = new Date()` likely use local TZ. Maybe the parsing of ISO date from API already takes all that into account and all is good but figured it would be worth pondering and checking.

AliSoftware (commented):
Skimmed the diff and LGTM 👍

Quite happy to have this feature and kind of milestone warning implemented 👌

_I'm a bit disappointed that apparently in TypeScript we need to instanciate a Date() only to setTime() on it, so creating a date object requires 3 lines (let, = Date(), setTime) and not just one? 😛 Wonder if there isn't a better way… But eh 🤷‍♂️ as long as it works…)_

Just left a question about TZ (maybe worth having a test covering that?) otherwise looks nice 🙂

org/pr/milestone.ts
mokagio: Anyone in the @Automattic/type-review team have any suggestion regarding this?

Not directly specific to type annotation, but hopefully you TypeScript pros have some good suggestions for us 🙇‍♂️

sirreal: This doesn't seem specific to TypeScript, but to JavaScript and Dates.

If I understand, we're parsing an ISO 8601 date string which may or may not include an offset. If it's coming from the API, I'd bet it does and it's UTC, but that seems outside our concern. If it's ISO 8601, we should be able to parse it correctly with [`Date.parse()`](https://developer.mozilla.org/en-US/docs/Web/Java;Script/Reference/Global_Objects/Date/parse) into a number representing the "number of milliseconds since January 1, 1970, 00:00:00 UTC." We do something similar with the current date ([`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) would have been a good alternative for that).

Using [`Date.getTime()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime):

> getTime() always uses UTC for time representation. For example, a client browser in one timezone, getTime() will be the same as a client browser in any other timezone.

So, the milliseconds are compared to see if the difference is less than 4 days, we're comparing UTC-based millisecond offsets (apples vs. apples).

Seems fine to me.

AliSoftware: Thanks for confirming! 🎉

PR #68, opened 2021-02-19
Avoid warning about soon to close milestones on merged PRs

loremattei:
This PR fixes an issue with the `milestone` rule where the warning about having set a milestone which is close to its due date is triggered also on merges. 

Looking at [GitHub documentation](https://docs.github.com/en/rest/reference/pulls), the state of a PR is set to `closed` when a PR is closed or merged, so checking for `state == "merged"` should cover both cases.

### DIFF
--- a/tests/pr-milestone.test.ts
+++ b/tests/pr-milestone.test.ts
@@ -145,4 +145,16 @@ describe("PR milestone checks", () => {
             expect(dm.warn).not.toHaveBeenCalled();
         })
     }
+
+    it("does not warn the milestone is closing in 4 days or less and PR is merged", async () => {
+        var closeDate = new Date();
+        closeDate.setDate(closeDate.getDate() + 4);
+
+        const mockData = { data: { draft: false, milestone: { number: 1, due_on: closeDate.toISOString() }, state: "closed" } };
+        dm.danger.github.api.pulls.get.mockReturnValueOnce(Promise.resolve(mockData));
+
+        await milestone();
+
+        expect(dm.warn).not.toHaveBeenCalled();
+    })
 })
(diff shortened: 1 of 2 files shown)


tests/pr-milestone.test.ts
```diff
@@ -145,4 +145,16 @@ describe("PR milestone checks", () => {
             expect(dm.warn).not.toHaveBeenCalled();
         })
     }
+
+    it("does not warn the milestone is closing in 4 days or less and PR is merged", async () => {
```
AliSoftware, line 149:
```suggestion
    it("does not warn when the milestone is closing in 4 days or less but PR is merged", async () => {
```

tests/pr-milestone.test.ts
```diff
@@ -145,4 +145,16 @@ describe("PR milestone checks", () => {
...
+        var closeDate = new Date();
+        closeDate.setDate(closeDate.getDate() + 4);
```
AliSoftware, line 151: Just a suggestion to avoid using the value exactly at the limit value `4` that we use between "closing soon" and "not closing soon" (in risk of hitting edge case of `<` vs `≤` unexpectedly and making the test actually test "not closing soon" by mistake – even if in that case that shouldn't change anything).

```suggestion
        closeDate.setDate(closeDate.getDate() + 2);
```

Just reduces the risk of doubt when reading the test code imho; but not an issue per se, so feel free to ignore 🤷

AliSoftware (approved):
Thanks for the fix!
Just left a couple of nits but nothing blocking, so LGTM! 👍

loremattei:
Thank you @AliSoftware! 🙇‍♂️  I integrated your suggestions and I'm merging this now.

PR #70, opened 2021-05-05
Remove versionName change rule for WordPress-Utils-Android

mokagio:
This should not be required anymore, see https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/76#issuecomment-832426470 .

I'm leaving this to draft with the same rationale as https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/77 .


oguzkocer (approved):
@mokagio It looks like this was the only `android-library.ts` usage and considering we'll change every Android library not to rely on the `versionName` property value, how do you feel about removing this file and its tests?

mokagio:
Good point @oguzkocer! Done in `8358596`, will merge on green.

PR #71, opened 2021-05-06
Update milestone check after "Part of a WIP Feature" label change

designsimply:
The `Part of a WIP Feature` label was changed to `status: do not merge` in the `woocommerce-android` and `woocommerce-ios` repos.

Since the milestone check Peril rule is used for other repos which may still have the `Part of a WIP Feature`, I tried making it so the update will still work if either label is used. I am actually not sure if I did it right or if I should simply have copied the entire block and change out the constant and label name. Does that make sense?

<sup>(internal reference: paaHJt-244-p2)</sup>

### DIFF
--- a/org/pr/milestone.ts
+++ b/org/pr/milestone.ts
@@ -22,7 +22,8 @@ export default async () => {
 
         // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
         const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
-        if (!targetsDevelop && !targetsRelease && wipFeature) {
+        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: do not merge"));
+        if (!targetsDevelop && !targetsRelease && (wipFeature || donotmergeFeature)) {
             return;
         }
     }
--- a/org/pr/milestone.ts
+++ b/org/pr/milestone.ts
@@ -22,7 +22,7 @@ export default async () => {
 
         // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
         const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
-        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: do not merge"));
+        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: feature-flagged"));
         if (!targetsDevelop && !targetsRelease && (wipFeature || donotmergeFeature)) {
             return;
         }


oguzkocer (changes requested):
@designsimply I don't think `status: do not merge` is a proper replacement for `Part of a WIP Feature`. The goal for the `Part of a WIP Feature` label is to give developers a way to skip some peril checks for PRs that target a WIP branch.

Is there a particular reason we don't want to keep the `Part of a WIP Feature` label along with `status: do not merge` label?

AliSoftware:
> @designsimply I don't think `status: do not merge` is a proper replacement for `Part of a WIP Feature`. The goal for the `Part of a WIP Feature` label is to give developers a way to skip some peril checks for PRs that target a WIP branch.

I agree with Oguz that while it made sense to allow not having a milestone when the label is named "Part of a WIP feature" , it doesn't really make sense to allow not having a milestone just because an issue is "status: do not merge".

If anything, amongst the new labels that you're putting in place and proposed in p91TBi-52p-p2, the one that could fit that bill imho would instead probably be `status: feature-flagged`, since the PRs and features that are feature-flagged – in order to not be included in release builds despite being merged on `develop`, because they will need multiple sprints and iterations before being ready to be released to the public – are closer to the meaning we gave to the WIP label, and its reason why we would allow not having milestones for those spread-across-multiple-sprints PRs.


> Is there a particular reason we don't want to keep the `Part of a WIP Feature` label along with `status: do not merge` label?

@oguzkocer correct me if I'm wrong but according to my reading of the diff, Sheri _did_ keep both labels in that check, to support both at once.

designsimply:
> Is there a particular reason we don't want to keep the `Part of a WIP Feature` label along with `status: do not merge` label?

Oof, no. They are both still there (as Olivier noted 👍) but with different names and now that you've asked about it I'm pretty sure I've done the Peril update for this one incorrectly. Here's what got updated: 

* Changed `Part of a WIP Feature` to `status: feature-flagged`.
* Changed `[Status] Not Ready for Merge` to `status: do not merge`.

So I think I should have changed the rule in the milestone check ([pr/milestone.ts](https://github.com/Automattic/peril-settings/blob/master/org/pr/milestone.ts)) to `status: feature-flagged` instead of `status: do not merge`. That's my mistake, so I will fix it. But first…

I have another question for you, would you prefer a different name for that label? Olivier also expressed concern about the naming for this label and you're the 2nd person to raise the issue so I can see that, in this case, I was trying to be extra helpful and failed. 🙂 But I think we can fix it! For label naming, I'm looking to do key: value pairs, all lower case, and trying to avoid inside-speak or acronyms (if possible!). To keep in line with the previous naming but also still do key: value and lower-case, do you think it would be better to change `status: feature-flagged` to `status: wip feature`? Or we could change it back to `Part of a WIP Feature` and just leave that one outside the other naming conventions for now to make it easier and re-visit when we update label names for the WordPress projects (this might be best now that I understand better how the Peril checks are used across so many different projects).

To summarize, which option?

* Option 1 - change `status: feature-flagged` to `status: wip feature` (or similar) and use it in the Peril milestone check.
* Option 2 - change `status: feature-flagged` back to the original `Part of a WIP Feature` (for now) and abandon this Peril update.

oguzkocer:
> @oguzkocer correct me if I'm wrong but according to my reading of the diff, Sheri did keep both labels in that check, to support both at once.

@designsimply @AliSoftware The issue I am trying to raise is not with the changes in this PR, but rather this part of the PR description:

> The `Part of a WIP Feature` label was changed to `status: do not merge` in the `woocommerce-android` and `woocommerce-ios` repos.

We have made that label available in all our major repos and made an internal announcement for it with the following information:


> ### Disabling milestone check for draft PRs and PRs against feature branches

> > It can be difficult to estimate the eventual milestone for draft PRs and PRs against long running feature branches. To address this, we are disabling the milestone check on draft PRs and PRs with the following label: `Part of a WIP Feature`.

We even came up with the name of this label together with @designsimply in a DM conversation. So, my feeling is that renaming/removing this label shouldn't be done without a good reason (not saying there isn't one, but rather asking if there is one) and without announcing this change.

What do you think?

designsimply:
Gotcha. The reason for changing the label is part of a larger set of projects to synchronize labels to a more a standard set across our repositories with idea that it will provide an easier onboarding experience and less friction for existing code wranglers when working across multiple code repositories, and it will help with tracking metrics across repositories too. A new naming convention was proposed, discussed, and approved using the following principles:

> Labels should be explicit, clear, and concise.
> Labels should be lowercase words separated by space.
> Key-value pairs are encouraged as they provide more context, e.g. type: bug.
> Superfluous labels should be avoided.
> Always fill in the description of labels.
> Use colors wisely, taking accessibility into consideration.
> Less is more. 
> Keep iterating. 🙂

This project is part of the Platform9 2021 plans and was first officially published on Jan 4th as part of our planning documentation in paaHJt-1Gt-p2—the plan name is "GitHub label unification / documentation" and the plan description is to: clean up and organize our GitHub labels across all app repositories to use a consistent core set of labels everywhere. A proposal was published on Apr 7th in p91TBi-4Rd-p2 and we discussed the `Part of the WIP Feature` in several comments on that post. Updates were published on Apr 29th at p91TBi-52p-p2 with additional changes based on the previous discussion. The label principles noted above are from a previous discussion done by the Woo core team to synchronize GitHub labels across repositories last year in several posts (see p6q8Tx-1Sy-p2 from Oct 2020).

I apologize for not including more information about the whys or references back to those things in this PR earlier. Thanks for your questions and please let me know I've missed anything and if you have any strong feelings about option 1 or 2 from my comment above now that you know a bit more about the project and it's background.

oguzkocer:
It looks like I crashed the party on this one. Since you had a long conversation about this, I trust that you have already weighed pros & cons, so I'll leave it to you.

org/pr/milestone.ts
```diff
@@ -22,7 +22,8 @@ export default async () => {
 
         // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
         const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
-        if (!targetsDevelop && !targetsRelease && wipFeature) {
+        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: do not merge"));
```
AliSoftware, line 25:
```suggestion
        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: feature-flagged"));
```
Or whatever you end up deciding renaming that label after your previous comment.

> I have another question for you, would you prefer a different name for that label? Olivier also expressed concern about the naming for this label and you're the 2nd person to raise the issue

Regarding that btw, I'm fine to keep `status: feature-flagged` or to pick a name that would convey the meaning that it's "a feature that is developed in incremental steps across multiple sprints" and "is currently not visible to end user in release builds so if there's a milestone set it's just an estimate and it's ok to push it to next sprint if not ready in time for this one". Since that's the whole meaning I want us to give to that label, and also the whole point of this Peril check here (if it's such a feature with the characteristics I wrote in quotes above, then it's the only case where it's ok to have a milestone on the PR).

AliSoftware (changes requested):

designsimply:
I appreciate the discussion on this PR! I really argued against keeping the label as `Part of a WIP Feature` in earlier discussions because I wanted to move to the new naming conventions and talking it over with both of you here helped a lot with context and for me learning more about how to make sure some announcements get the right reach. 

I feel okay to use the new label on this one to try it out but be open to either switching it back or updating it as we begin to iterate on the other repos later.

Thanks so much!

org/pr/milestone.ts
```diff
@@ -22,7 +22,8 @@ export default async () => {
 
         // Skip for PRs for wip features unless the PR is against "develop" or "release/x.x" branches
         const wipFeature = githubLabels.some(label => label.name.includes("Part of a WIP Feature"));
-        if (!targetsDevelop && !targetsRelease && wipFeature) {
+        const donotmergeFeature = githubLabels.some(label => label.name.includes("status: feature-flagged"));
+        if (!targetsDevelop && !targetsRelease && (wipFeature || donotmergeFeature)) {
```
AliSoftware, line 26:
```suggestion
        const featureFlagged = githubLabels.some(label => label.name.includes("status: feature-flagged"));
        if (!targetsDevelop && !targetsRelease && (wipFeature || featureFlagged)) {

```

AliSoftware:
Also, you might want to also modify the unit tests in `tests/pr-milestone.test.ts` in order to add some tests about this new label name.

I suggest you just duplicate the existing tests that make use of `name: 'Part of a WIP Feature'` for the `dm.danger.github.issue.labels` fake, using the name of the label in those duplicated test cases with the new label.

PR #72, opened 2021-05-06
Update tracks check after "Tracks" label change

designsimply:
The `Tracks` label was changed to `category: analytics` in the `woocommerce-android` and `woocommerce-ios` repos.

Since the Peril rule is used for other repos as well, I updated the comment to ask for people to assign the proper label depending on what repo they are using. I couldn't figure out if there are any other implications for the `Tracks` label but I did notice that there were very few PRs using that label at all in the woocommerce repos before the label was updated.

<sup>(internal reference: paaHJt-244-p2)</sup>

### DIFF
--- a/org/pr/check-tracks.ts
+++ b/org/pr/check-tracks.ts
@@ -68,7 +68,7 @@ export default async () => {
         // Put it all together! 
         messageText = `This PR contains changes to Tracks-related logic. Please ensure the following are completed:\n`;
         messageText += `**PR Author**\n`;
-        messageText += `- The PR must be assigned the **Tracks** label\n`;
+        messageText += `- The PR must be assigned the label **Tracks** or **category: analytics**, depending on which label is avail in the repo you are using.\n`;
         messageText += `**PR Reviewer**\n`;
         messageText += `- The tracks events must be validated in the Tracks system.\n`;
         messageText += `- Verify the internal tracks spreadsheet has also been updated.`;


org/pr/check-tracks.ts
```diff
@@ -68,12 +68,12 @@ export default async () => {
         // Put it all together! 
         messageText = `This PR contains changes to Tracks-related logic. Please ensure the following are completed:\n`;
         messageText += `**PR Author**\n`;
-        messageText += `- The PR must be assigned the **Tracks** label\n`;
+        messageText += `- The PR must be assigned the label **Tracks** or **category: analytics**, depending on which label is avail in the repo you are using.\n`;
```
AliSoftware, line 71:
```suggestion
        messageText += `- The PR must be assigned the label **Tracks** or **category: analytics**, depending on which label is available in the repo you are using.\n`;
```

To be honest I'm surprised we didn't go further in our logic when implementing that rule and make the rule check for those labels, instead of making a message asking the author to check that.

It could be nice to take this occasion to add such a check, something like this (not tested, bight have syntax errors, but you get the gist):

```ts
const hasTracksLabel = githubLabels.some(label => label.name.includes("Tracks") || label.name.includes("category: analytics"));
messageText = `This PR contains changes to Tracks-related logic. Please ensure the following are completed:\n`;
if (!hasTracksLabel) {
  messageText += `**PR Author**'n`;
  messageText += `- The PR must be assigned the **Tracks** or **category: analytics** label – depending on which label is available in the repo you are using.\n`;
}
messageText += `**PR Reviewer**\n`
…
```

designsimply: > To be honest I'm surprised we didn't go further in our logic when implementing that rule.

Maybe it's not needed? Is there a reason it must be strictly required?

fwiw, this is the one spot where I wasn't sure why the suggestion was there or how the `Tracks` label is used in automation elsewhere, if at all. I thought that perhaps it was just there as a reminder for organizational purposes but am also curious about it.

AliSoftware: > Is there a reason it must be strictly required?

To be honest I have no idea 😅

> fwiw, this is the one spot where I wasn't sure why the suggestion was there or how the `Tracks` label is used in automation elsewhere, if at all. 

You and me both. I have to say I'm not familiar with how we handle that and what's the prices around these (and I'm on my phone so I can't MGS for it). Maybe @jkmassel has better context?

AliSoftware: Ping @jkmassel about the questions above.

If there's no concerns there tho, I think we can merge this (cc @designsimply); I just wanted to get your PoV about that whole Tracks process before to check if we needed to adapt the message while we're at it.

PR #76, opened 2021-05-12
Fix installable build comments on custom statuses

loremattei:
With the recent changes related to the Jetpack app, Peril stopped to update the `Installable Build` comment on PRs. The root cause of the issue is that the job name used for the WordPress and Jetpack installable builds now contains the app name and, so, the status reported by CircleCI doesn't match the expected one anymore. 
Peril expects a `success` status with `ci/circleci: Installable Build` as the context, but the new context is either `"ci/circleci: WordPress Installable Build"` or `"ci/circleci: Jetpack Installable Build"`.

This PR fixes the issue by using regex for the expected statuses.

### DIFF
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -9,7 +9,9 @@ const PERIL_BOT_USER_ID: number = parseInt(process.env['PERIL_BOT_USER_ID'], 10)
 
 // This is a list of the CircleCI statuses to process
 const HOLD_CONTEXTS: string[] = ["ci/circleci: Installable Build/Hold", "ci/circleci: Installable Build/Approve WordPress"]
-const INSTALLABLE_BUILD_CONTEXTS: string[] = ["ci/circleci: Installable Build", "ci/circleci: Test Android on Device"]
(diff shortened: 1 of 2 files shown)


org/pr/installable-build.ts
```diff
@@ -131,10 +133,11 @@ async function getDownloadCommentText(status) {
 }
 
 export default async (status: Status) => {
+  var pippo = INSTALLABLE_BUILD_CONTEXTS.filter(s => status.context.match(s));
     if (status.state == "pending" && HOLD_CONTEXTS.includes(status.context)) {
```
AliSoftware, line 137: Seems like leftover from debugging session there 🙃

(TIL about the `debugger` command, seems useful!)

AliSoftware (approved):
LGTM… as long as you remove the leftover debugging lines before merging 😛

org/pr/installable-build.ts
loremattei: Good catch! 🤦‍♂️ Clearly trying to do too many things in parallel.

PR #78, opened 2021-06-07
Remove subtree checks

oguzkocer:
With https://github.com/wordpress-mobile/WordPress-Android/pull/14742 & https://github.com/woocommerce/woocommerce-android/pull/4108 we won't have login subtree in our clients, so this check is not necessary anymore.

@loremattei Couple unrelated notes:
1. We seem to be still using `master` as our branch, should we switch that to `trunk`?
2. Current develop seems to be [failing in CI for `Restart automatic-peril` task](https://app.circleci.com/pipelines/github/Automattic/peril-settings/213/workflows/5d6a97cf-3bb4-4de6-9bf7-0fe1aca91814/jobs/284)


loremattei:
Great!

I think we want to remove the calls to it [here](https://github.com/Automattic/peril-settings/blob/e770e8c7bd947f34748ae4ca34064414ddbe452d/peril-settings.json#L32) and [here](https://github.com/Automattic/peril-settings/blob/e770e8c7bd947f34748ae4ca34064414ddbe452d/peril-settings.json#L74) as well. 

Other than that, it looks good to me :-) 

> 1. We seem to be still using master as our branch, should we switch that to trunk?

Last time I checked, Peril was not able to pick the rules from anything other than `master`. It's been a while tho, so we should probably check again and switch to `trunk` if possible 👍

> 2. Current develop seems to be failing in CI for Restart automatic-peril task

@jkmassel can this be related to the changes you made on Heroku a few weeks ago?

oguzkocer:
> I think we want to remove the calls to it here and here as well.

Fixed in 81e41631983e501c539a049f5cff3adfe2426481

> Last time I checked, Peril was not able to pick the rules from anything other than master. It's been a while tho, so we should probably check again and switch to trunk if possible 👍

Oh, that makes sense 👍 

This is ready for another look @loremattei 🙇

loremattei (approved):
LGTM!

PR #79, opened 2021-06-25
Handle multiple APKs installable for an Installable Build

AliSoftware:
The link in the Peril comment for the installable WPAndroid build (e.g https://github.com/wordpress-mobile/WordPress-Android/pull/14936#issuecomment-868401819) is actually pointing to the Jetpack app.

We're able to download the apk if we visit the artifacts page anyway, but it's nicer to have Peril provide the link for all installable APKs not just the first one.

\cc @hypest

### DIFF
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -106,7 +106,7 @@ async function circleCIArtifacts(status) {
 }
 
 async function getDownloadCommentText(status) {
-  const artifacts = await circleCIArtifacts(status)
+  const artifacts: Array<any> = await circleCIArtifacts(status)
 
   const commentJsonArtifact = artifacts.find(artifact => artifact.path.endsWith("comment.json"))
   if (commentJsonArtifact) {
@@ -123,9 +123,12 @@ async function getDownloadCommentText(status) {
     }
   }
 
-  const apkArtifact = artifacts.find(artifact => artifact.path.endsWith(".apk"))
-  if (apkArtifact) {
-    return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifact.url}).`
+  const apkArtifacts: Array<any> = artifacts.filter(artifact => artifact.path.endsWith(".apk"))
+  if (apkArtifacts.length == 1) {
+    return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifacts[0].url}).`
+  } else if (apkArtifacts.length > 1) {
+    const links = apkArtifacts.map(artifact => `- [${artifact.path}](${artifact.url})`).join(`\n`)
+    return `You can test the changes on this Pull Request by downloading the APKs:\n${links}.`
   }
   return undefined
 }
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -127,7 +127,7 @@ async function getDownloadCommentText(status) {
   if (apkArtifacts.length == 1) {
     return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifacts[0].url}).`
   } else if (apkArtifacts.length > 1) {
-    const links = apkArtifacts.map(artifact => `- [${artifact.path}](${artifact.url})`).join(`\n`)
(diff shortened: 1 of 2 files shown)


tests/installable-build-test.ts
```diff
@@ -151,4 +151,29 @@ describe("installable build handling", () => {
...
+        target_url: "https://circleci.com/gh/Owner/Repo/12345?some=query",
+        repository: {
+            name: 'Repo',
+            owner: { login: 'Owner' }
+        },
+        commit: { sha: 'abc' }
+      }
+      await installableBuild(webhook)
+
+      expectComment(webhook, `You can test the changes on this Pull Request by downloading the APKs:\n- [file1.apk](${mockedArtifacts[0].url})\n- [file2.apk](${mockedArtifacts[1].url}).`)
```
AliSoftware, line 177: 🤔 I'm hoping that `\n` are correctly handled in that case once posted as a GitHub comment – as opposed to having to use `<br/>` to ensure we go to newline?

hypest:
👋 @AliSoftware ! I wonder, can you share some testing instructions to verify if/how the changes here work? I'm not familiar with the Peril integration and I wonder how can I go about reviewing this PR. Thanks!

AliSoftware:
The testing instructions are a bit tricky for Peril, because it's written in TypeScript… so if you want to run the unit tests (which I updated in that PR), you need to have `npm` and `yarn` installed, and then you can run `yarn test` to run the unit tests.

But this is something that the CI already does (running unit tests), so if the CI is green that means those passed and no need to test locally.

I don't think we can test much more (e.g testing the rule on an actual PR) easily given this is a bit that acts on all our repos and only uses the config and rule files in `master`, not in PRs (I mean, [there are ways to run Peril locally either on your mac or in Docker](https://github.com/danger/peril/blob/master/docs/local_dev.md), and then point it to a PR and all, but way too much trouble for what it's worth, so we never do that and just trust our unit tests.)

hypest:
Thanks for elaborating Olivier! I'll give the diff a visual look and probably optimistically give the go ahead if nothing captures my attention :)

tests/installable-build-test.ts
```diff
@@ -151,4 +151,29 @@ describe("installable build handling", () => {
 
       expectComment(webhook, `You can test the changes on this Pull Request by downloading the APK [here](${mockedArtifacts[0].url}).`)
     })
+
+    it("Posts a download comment linking to multiple APKs", async () => {
```
AliSoftware, line 155: Note that this unit test is just a shameless copy/paste of the previous unit test (lines 133-153) for single-APK case, adjusted to have a mockedArtifacts containing multiple APKs and a adjusted expectation.

org/pr/installable-build.ts
```diff
@@ -123,9 +123,12 @@ async function getDownloadCommentText(status) {
...
 
-  const apkArtifact = artifacts.find(artifact => artifact.path.endsWith(".apk"))
-  if (apkArtifact) {
-    return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifact.url}).`
+  const apkArtifacts: Array<any> = artifacts.filter(artifact => artifact.path.endsWith(".apk"))
+  if (apkArtifacts.length == 1) {
+    return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifacts[0].url}).`
+  } else if (apkArtifacts.length > 1) {
+    const links = apkArtifacts.map(artifact => `- [${artifact.path.split("/").pop()}](${artifact.url})`).join(`\n`)
+    return `You can test the changes on this Pull Request by downloading the APKs:\n${links}.`
```
hypest, line 131: Nit: wdyt about removing the `.` at the end since I kinda feel it will look weird when we have multiple lines and only the last one has the period?

hypest (approved):
LGTM @AliSoftware ! I only left a very minor comment, otherwise looks good to me.

While at it I verified that tests do run correct locally as well 👍

org/pr/installable-build.ts
AliSoftware: Yep, this was an overlook after copy/pasting from the wording of the single-APK case. Will fix, thanks!

PR #80, opened 2021-07-14
Update/diff size rule

loremattei:
This PR updates the Max Diff size rule for Android. 

Previously, we raised a warning on PRs which contained more than 500 lines of changes. Android developers pointed out that 500 lines of actual code changes can already be too many and that, on the other end, a few hundreds of lines of changes may require quite a lot of changes in the tests. 

Because of this, this PR:
 - reduces the maximum number of lines of code changes to 300. 
 - removes the changes to files in `/src/test` from the calculation. 

This PR doesn't affect the behaviour for iOS projects.


org/pr/android-diff-size.ts
```diff
@@ -0,0 +1,30 @@
...
+    for (let file of modifiedStrings) {        
+        const stringDiffs = await danger.git.diffForFile(file)
+        const addedLines = stringDiffs.added.split("\n").filter(x => x.startsWith("+")).length;
+        const removedLines = stringDiffs.removed.split("\n").filter(x => x.startsWith("-")).length;
+        changesToTests += addedLines + removedLines;
+    }
+
+    // Warn when there is a big PR
+    let codeChanges = danger.github.pr.additions + danger.github.pr.deletions - changesToTests;
+    if (codeChanges > 300) {
```
AliSoftware, line 27: I'm curious to what the rationale/benefit is to "compute the diff size for tests, then substract that from the `pr.additions+pr.deletions` total") vs directly computing the diff size for non-tests?

In other words, why not directly run your code from lines 18-23 on the non-test files?

```suggestion
    // Calculate the changes to non-test files because we don't want to limit those
    const nonTestFiles = modifiedFiles.filter((path: string) => !path.includes("/src/test"));
    let codeChanges = 0;
    for (let file of nonTestFiles) {        
        const stringDiffs = await danger.git.diffForFile(file)
        const addedLines = stringDiffs.added.split("\n").filter(x => x.startsWith("+")).length;
        const removedLines = stringDiffs.removed.split("\n").filter(x => x.startsWith("-")).length;
        codeChanges += addedLines + removedLines;
    }

    // Warn when there is a big PR
    if (codeChanges > 300) {
```

tests/android-diff-size.test.ts
```diff
@@ -0,0 +1,135 @@
...
+            },
+            issue: {
+                labels: []
+            },
+        },
+    };
+});
+
+describe("PR diff size checks", () => {
+    it("does not warn with less than 300 lins", async () => {
```
AliSoftware, line 30: 💄 
```suggestion
    it("does not warn with less than 300 lines", async () => {
```

tests/android-diff-size.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        await diffSize();
+        
+        expect(dm.warn).not.toHaveBeenCalled();
+    })
+
+    it("warns with more than 300 lines", async () => {
+        dm.danger.github.pr.deletions = 101;
```
AliSoftware, line 37: Completely opinionated so feel free to ignore, but I wonder if it wouldn't be easier to read if the mocked data were 0,0 and you did set both `pr.additions` and `pr.deletions`'s values in each test?

Because I almost commented saying "how does this test pass given your limit is 300 but you only use 101 which is less than 300" and it took me a second to realize that `pr.additions` had a default value of 200 in the mocks and that's where it all came from.

tests/android-diff-size.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        dm.danger.github.pr.additions = 291;
+        dm.danger.github.pr.deletions = 12;
+
+        dm.danger.git.modified_files = ["MyProject/src/test/something/myTestFile.ko"];
+        dm.danger.git.diffForFile = jest.fn();
+
+        const mockData = { 
+            added: "src/test/something/myTestFile.ko\r\n+ func someNewCode();\r\n- let someOtherNewVar = 0;",
+            removed: "src/test/something/myTestFile.ko\r\n- func someCode();\r\n- let someOtherVar = 0;" 
+        };
```
AliSoftware, line 104: Are we sure that's the format of the structure returned by Danger and the git gem?

I'm confused as to what changes would go in `added:` and which ones would go in `removed:` keys in those cases. For example here in `added` you have a diff that contains both `+` and `-` lines, so isn't that a mix of both an addition and a deletion, and as such why would it be in `added:` rather than `removed:`

I am not sure what the `dm.danger.git.diffForFile` function actually returns (when not mocked) in those kind of cases, and how it differently represent new files vs deleted files vs modified files that contains only deleted lines, vs modified files that contains only added lines, vs modified files that might contain a mix of additions and deletions… but I think it's worth double-checking here (otherwise this test wouldn't actually test the actual case we'd encounter in real life)

org/pr/android-diff-size.ts
loremattei: Oh, this is just for speed. The values of total additions and deletions are returned by the API and, in many PRs, they are all we need.

tests/android-diff-size.test.ts
loremattei: Oh, I see! I find it slightly easier to have a default and have only the diff for that particular test in the test code, but I'm totally fine with both. 
I think that most of tests in this repository are organized this way though, so probably we should keep it for consistency.

tests/android-diff-size.test.ts
loremattei: Good eye :-)

tests/android-diff-size.test.ts
loremattei: Oh, right! That's a typo. There should be only `+` in `added` and only `-` in `removed`.

AliSoftware: Ah! Makes more sense now 👍 

I dug a little and found [the definition GitDSL from Danger](https://github.com/danger/danger-js/blob/main/source/dsl/GitDSL.ts#L7-L18), and it seems like `added` and `removed` are guaranteed to only contain added and removed lines. Which means that we could probably just count all the lines in those strings, without having to filter by `+` or `-`. In other words, you should be able to get rid of [the `filter(…)` calls on lines 20-21](https://github.com/Automattic/peril-settings/pull/80/files#diff-4ad737252bc200d7661627f749fe3cac6d9023a8ea69a835236657fb40c29bf9R20-R21) and avoid useless CPI cycles filtering those lines 😉

org/pr/android-diff-size.ts
AliSoftware: Ah ok, so you estimated that there will be more non-test files than test files so it was quicker, got it!

That being said, while looking at the Danger docs to dig a bit deeper into the structure of the `diffForFile` return value, I also found [this `linesOfCode` method](https://github.com/danger/danger-js/blob/main/source/dsl/GitDSL.ts#L191) in the GitDSL exposed to Danger… which actually seems to do everything we want here – and actually a bit better, since it [also considers created files and deleted files in the count](https://github.com/danger/danger-js/blob/main/source/platforms/git/gitJSONToGitDSL.ts#L160-L184), which apparently we forgot to include here as we only consider `modifiedFiles`. I think it could allow us to simplify the code quite a lot here!🎉

AliSoftware: If I'm not mistaken that means that you could replace all the lines 15-23 with just `changesToTests = danger.git.linesOfCode('/src/test/**')`. I'm not sure if there's a way to negate a glob pattern (to directly get `codeChanges = danger.git.linesOfCode('!/src/test/**')` or something, assuming that's supported), but even use the positive glob then substract to the total would be both more accurate and readable than what we initially tried here, right?

AliSoftware: Ooooh apparently negative glob _is_ supported, since the [DangerJS code seems to rely](https://github.com/danger/danger-js/blob/main/source/platforms/git/gitJSONToGitDSL.ts#L161) on the [micromatch library](https://github.com/micromatch/micromatch) and in the README of that lib they have the following example:

```
console.log(micromatch(['foo', 'bar', 'baz', 'qux'], ['*', '!b*'])) //=> ['foo', 'qux']
```

I say it's worth a try, since I'd allow us to make our whole code logic down to a single `linesOfCode('!/src/test/**')` if that indeed works!

loremattei: Yup, it was great if we could use that API. Unluckily this has been implemented in a newer version of Danger than the one currently integrated in Peril. The version of the API we have available actually implements `linesOfCode`, but it is an older version which doesn't accept any parameter for filters yet :-(

There's an [open issue related to updating Danger in Peril](https://github.com/danger/peril/issues/470), but it seems nobody is currently actively working on it. It could be a nice HackWeek project :-)

tests/android-diff-size.test.ts
loremattei: Yup! I initially did it. But during my debug, I found that sometimes the file name doesn't get parsed out and so it can be counted as an additional line if we don't apply a proper filtering. I believe this is a bug (it doesn't happen always and it's different from the documentation), so it's possible that it's fixed in a newer version of Danger, but as per my comment above, we are using a quite older version.

org/pr/android-diff-size.ts
AliSoftware: Ah, good catch. Dammit 😒  Let's keep that manual then (though it could be nice to add a comment with a link to that issue so we consider updating it once it's solved and shipped).

---

What do you think of including `createdFiles` and `deletedFiles` in addition to `modifiedFiles` when counting lines in tests? (This would probably look like this below but haven't tested if that's valid syntax). I'm especially thinking about the case that most PRs that currently a high count because of tests (that we now want to ignore) could be from new features, and thus potentially introducing new test files alongside the new feature, so that might actually be one of the most common case of the highest line counts due to tests in PRs…

```
const allFiles = (modifiedFiles + createdFiles + deletedFiles).filter((path: string) => path.includes("/src/test"));
```
Just hoping that `diffForFile(aDeletedFile)` is valid (and returns all its lines in the `removed:` key) and wouldn't e.g. throw an error for file not found, but I think that would work.

loremattei: Btw, while looking briefly into that update, I managed to implement branch support for the configuration file. Danger already has it, but Peril doesn't. I'll open a PR soon as I think it will make testing changes easier.

tests/android-diff-size.test.ts
AliSoftware: Ah, thanks for the explanation; good workaround then 👍

org/pr/android-diff-size.ts
AliSoftware: image

loremattei: > What do you think of including createdFiles and deletedFiles in addition to modifiedFiles when counting lines in tests?

Ah, good point! Thanks for catching it. Added in `0b507cf`

AliSoftware (approved):
:shipit: 🚀

PR #81, opened 2021-07-14
Update optional tests state for Simplenote-iOS

loremattei:
While working [on a small PR on Simplenote iOS](https://github.com/Automattic/simplenote-ios/pull/1318) I noticed that the status of Optional Tests which are on hold is not green. 

This PR adds the Simplenote iOS Optional Tests to the list of jobs which PR updates to avoid that the status of the PR never goes green.


oguzkocer (approved):
Thank you for this @loremattei!

PR #84, opened 2021-08-18
Update peril-settings to reduce GitHub API calls

AliSoftware:
### Why

While looking at p1629260121056500-slack-C02QANACA I noticed that we ran all our main Danger rules on _every_ action of the `pull_request` event, while some of them only make sense to be run on specific ones.

 - In particular, we run those rules even on actions like `pull_request.closed` while at that stage there's no need to re-run Danger.  
 - We also re-run rules that run checks on certain things that couldn't have been changed by certain events, e.g. we tested the labels even on new commits pushed on the PR without any label change, or re-checked the release notes or the diff size even just when the only thing that triggered the webhook was a label being added, etc.

As a result, we ended up having Danger/Peril failures from time to time, due:
 - Either to glitches in the API that made Danger fail when running some rules on PR closed (see Slack conversation linked above), while there was no point to re-run Danger in those cases in the first place
 - Or us reaching the GitHub API rate-limit, which when it happens makes all the subsequent API calls and all Danger/Peril runs on all pull requests fail for a couple of hours until the limit is cleared. (see p1623162838019800-slack-CEJT26EG1)

### How

By running some rules only for the specific events that makes sense for them, we should reduce the API usage and the risk of GitHub API and Peril erroring.

### DIFF
--- a/peril-settings.json
+++ b/peril-settings.json
@@ -8,15 +8,16 @@
     "rules": {},
     "repos": {
         "wordpress-mobile/WordPress-iOS": {
-            "pull_request, pull_request.labeled, pull_request.unlabeled": [
+            "pull_request": [
                 "Automattic/peril-settings@org/pr/ios-macos.ts",
-                "Automattic/peril-settings@org/pr/label.ts",
-                "Automattic/peril-settings@org/pr/milestone.ts",
+                "Automattic/peril-settings@org/pr/milestone.ts"
+            ],
+            "pull_request.opened, pull_request.labeled, pull_request.unlabeled, issues.opened, issues.labeled, issues.unlabeled": [
+                "Automattic/peril-settings@org/pr/label.ts"
+            ],
+            "pull_request.opened, pull_request.synchronize": [
                 "Automattic/peril-settings@org/pr/ios-diff-size.ts",
                 "Automattic/peril-settings@org/pr/release-notes.ts"
-            ],
-            "issues.opened, issues.labeled, issues.unlabeled": [
-                "Automattic/peril-settings@org/issue/label.ts"
             ],
             "status": [
                 "Automattic/peril-settings@org/pr/installable-build.ts",
@@ -24,15 +25,16 @@
             ]
         },
         "wordpress-mobile/WordPress-Android": {
-            "pull_request, pull_request.labeled, pull_request.unlabeled": [
+            "pull_request": [
                 "Automattic/peril-settings@org/pr/android.ts",
-                "Automattic/peril-settings@org/pr/label.ts",
-                "Automattic/peril-settings@org/pr/milestone.ts",
+                "Automattic/peril-settings@org/pr/milestone.ts"
+            ],
+            "pull_request.opened, pull_request.labeled, pull_request.unlabeled, issues.opened, issues.labeled, issues.unlabeled": [
+                "Automattic/peril-settings@org/issue/label.ts"
+            ],
+            "pull_request.opened, pull_request.synchronize": [
                 "Automattic/peril-settings@org/pr/android-diff-size.ts",
                 "Automattic/peril-settings@org/pr/release-notes.ts"
-            ],
-            "issues.opened, issues.labeled, issues.unlabeled": [
-                "Automattic/peril-settings@org/issue/label.ts"
             ],
             "status": [
                 "Automattic/peril-settings@org/pr/installable-build.ts",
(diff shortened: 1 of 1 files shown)


peril-settings.json
```diff
@@ -8,32 +8,34 @@
     "rules": {},
     "repos": {
         "wordpress-mobile/WordPress-iOS": {
-            "pull_request, pull_request.labeled, pull_request.unlabeled": [
+            "pull_request": [
```
AliSoftware, line 11: I'd love to reduce that filter even more, to only run the platform-specific rule (`ios-macos.ts` or `android.ts` depending on the repos) and the milestone rule only on events that makes sense. But:

## For the platform-specific rule

Since the rule is pretty generic (i.e. it contains many different checks), it feels safer to run it on every action. One would have to check the rule implementation in details to check what GitHub/Peril APIs we use on those rules (i.e. if we access the list of labels, the diff, etc) to know what filter would be appropriate, but given that rule is quite generic by nature and its implementation might evolve with new checks while not changing the rule name, it felt quite risky to tight it to only specific actions.

This could be discussed further, and we could at least consider only limiting to `pull_request.opened, pull_request.synchronize`, which would probably make sense (I'd be ok to change it to that, I'm just not sure we wouldn't miss any other action we wouldn't have thought of there), and would have the benefit to avoid `pull_request.closed` (which we definitively don't need there).

What do you think? If we limit to specific actions, what should be the full list for this rule so we don't forget any case?

## Milestone rule

I haven't found any action name for when a milestone is added. In fact, I tried to change the milestone of an opened PR, then went to look at [the delivered webhook payloads in GitHub](https://github.com/wordpress-mobile/WordPress-Android/settings/hooks/85494651/deliveries) and didn't see any event triggered from that milestone addition/removal. And the [list of actions for the `pull_request` event in GitHub API docs](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent) don't mention any action that could match the milestone addition either. This is why I ended up leaving this one under the generic `pull_request` without any specific action, as I don't know which one we should use if any.

In practice I think what's happening (even currently) is that adding/removing a milestone don't make Peril re-check for this rule (like we'd hope it does), and we need to do something else to the PR (like pushing new commits or adding labels) to force-retrigger it. So in this sense, this PR doesn't change that existing behavior.

Like for the platform rule, I'm opened to suggestions as to how we can improve that and which actions could be reasonable to filter on for this rule.

peril-settings.json
```diff
@@ -91,16 +105,24 @@
...
-                "Automattic/peril-settings@org/pr/label.ts",
+            "pull_request": [
+                "Automattic/peril-settings@org/pr/ios-macos.ts"
+            ],
+            "pull_request.opened, pull_request.labeled, pull_request.unlabeled, issues.opened, issues.labeled, issues.unlabeled": [
+                "Automattic/peril-settings@org/pr/label.ts"
+            ],
+            "pull_request.opened, pull_request.synchronize": [
                 "Automattic/peril-settings@org/pr/release-notes.ts"
             ]
```
AliSoftware, line 116: I noticed that simplenote-macos doesn't use neither the `milestone.ts` rule – that one might make sense given the SNmacOS app might not be on a regular schedule like our other apps? – but nor the `ios-diff-size.ts` rule. Is that intentional?

peril-settings.json
```diff
@@ -91,16 +105,24 @@
...
-                "Automattic/peril-settings@org/pr/label.ts",
+            "pull_request": [
+                "Automattic/peril-settings@org/pr/android.ts"
+            ],
+            "pull_request.opened, pull_request.labeled, pull_request.unlabeled, issues.opened, issues.labeled, issues.unlabeled": [
+                "Automattic/peril-settings@org/pr/label.ts"
+            ],
+            "pull_request.opened, pull_request.synchronize": [
                 "Automattic/peril-settings@org/pr/release-notes.ts"
             ],
```
AliSoftware, line 127: I was surprised to see that `simplenote-android` doesn't run the `milestone.ts` rule, nor the `android-diff-size.ts` rule. Is that intentional?

loremattei (approved):
LGTM!

I would add the missing rules to the other apps.

AliSoftware:
@loremattei Thx.
What's your take about the platform-specific rules and the milestone rules, should we keep them under `pull_requests` filter for now or do you think it's worth thinking about the appropriate, more restrictive filter for those?

> I would add the missing rules to the other apps.

👍 . I think I'll create a separate PR about the missing rules to other apps, just so I can add the relevant leads to it to confirm if that makes sense to them or not (e.g. confirm if `simplenote-macOS` uses milestones or not, etc) and move the discussion about these in that dedicated PR.

loremattei:
> What's your take about the platform-specific rules and the milestone rules, should we keep them under the catch-all pull_requests filter for now or do you think it's worth thinking about the appropriate, more restrictive filter for those?

For platform-specific rules I would keep them under the catch-all for now and, imho, we should split those platform rules into different files over the time to make it easier maintaining them and also applying the right filters. 

For milestone rules, it looks like you are right and they are not currently re-run when the milestone is added. IIRC it used to work as expected when we were monitoring the `issue.milestoned` property (or it was not a property? Maybe a state of `action`? I don't remember exactly). I don't know when/why we changed it tbh.

AliSoftware:
Ok, let's keep them both under catch-all for now then 👍

PR #85, opened 2021-08-18
Workaround to fix a glitch in Android diff size 

loremattei:
It looks like we are still experiencing the issue/bug/glitch where some properties sometimes become unavailable during the flow. This PR applies the workaround we found for this problem to the new Android Diff Size rule.

### DIFF
--- a/org/pr/android-diff-size.ts
+++ b/org/pr/android-diff-size.ts
@@ -6,6 +6,8 @@ export default async () => {
     const modifiedFiles = danger.git.modified_files;
     const createdFiles = danger.git.created_files;
     const deletedFiles = danger.git.deleted_files;
+    const additions = danger.github.pr.additions;
+    const deletions = danger.github.pr.deletions;
 
     if (githubLabels.length != 0) {
         const releases = githubLabels.some(label => label.name.includes("Releases"));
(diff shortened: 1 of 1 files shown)


org/pr/android-diff-size.ts
```diff
@@ -6,6 +6,8 @@ export default async () => {
     const modifiedFiles = danger.git.modified_files;
     const createdFiles = danger.git.created_files;
     const deletedFiles = danger.git.deleted_files;
+    const additions = danger.github.pr.additions;
+    const deletions = danger.github.pr.deletions;
```
AliSoftware, line 10: I think this could be worth a comment on line 5 before all those `const` to explain why this is important (i.e. workaround a glitch in Peril) and not just for style. And also to let others know that this Peril issue exist and that this is the workaround to solve it, for future us who were not aware and might need to apply it for other properties if that happens again in the future.

AliSoftware (approved):
I'd love to have a comment to mention the workaround, but otherwise code LGTM

org/pr/android-diff-size.ts
loremattei: Yup! Done in `c5e3b95`

AliSoftware: Perfect, thanks 👍

PR #86, opened 2021-08-23
Fix rule settings for labels

AliSoftware:
The recent PR https://github.com/Automattic/peril-settings/pull/84 accidentally merged the settings for `issue/labels.ts` and `pr/labels.ts` into a single set of events all related to Pull Requests only, resulting in the `Please add a type label to this issue. e.g. '[Type] Enhancement'` message from the `issue/labels.ts` rule now appearing in all `Releases` PRs (example [here](https://github.com/wordpress-mobile/WordPress-Android/pull/15228#issuecomment-903180621))

We thus missed the fact that there are actually two different rules for labels, one for issue and one for PR, and they still need to be kept on separate triggers.

---

In addition, I also:
 - added the `labeled`/`unlabeled` event triggers on the diff and release-notes rules as they both depends on which labels are set (and make exceptions for e.g. when the `Releases` label is set).
 - Applied the same workaround as #85 for `android.ts` – because I saw it fail on `danger.git.diffForFile(…)` in https://github.com/wordpress-mobile/WordPress-Android/pull/15233#issuecomment-903597527

### DIFF
--- a/org/pr/android.ts
+++ b/org/pr/android.ts
@@ -2,11 +2,15 @@ import {warn, danger} from "danger";
 
 export default async () => {
 
-    // If changes were made to the release notes in the metadata folder, there must also be changes to the PlayStoreStrings file.
+    // Store the relevant data
+    // This is a workaround for a weird issue/glitch we have been experiencing
+    // where, sometimes, the data is not accessible later in the flow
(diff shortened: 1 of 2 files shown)


loremattei (approved):
:shipit:

PR #90, opened 2021-09-06
Update the Installable Build job status for Jetpack

loremattei:
After https://github.com/Automattic/peril-settings/pull/89 got merged, I've started noticing that the status of the `Installable Build` job for the Jetpack application doesn't get updated from `pending (orange)` to `ready (green)`. 

The reason is that [this commit](https://github.com/Automattic/peril-settings/commit/1ed4cc4df9518037fc2287fd0278c15aace56c9d) removed it from the list of the handled states. 
My understanding, from the commit title, is that the goal was to avoid that Peril commented on the PR with the prompt to run the installable build for Jetpack.
I think we still want the state of the job to be green, because, otherwise, the PR remains in a pending state and, while it's still possible to merge the changes because the job is not required, this behaviour can create confusion. 
My take is that the bug that https://github.com/Automattic/peril-settings/pull/89 fixed was actually hiding this problem. 

This PR should fix this problem. There are probably better ways to do this, and the test cases are not complete and strong, but since @jkmassel mentioned that with the migration to Buildkite this function will be integrated in the CI and won't be required anymore in Peril, I opted for a quick, minimal and easy solution.

### DIFF
--- a/org/pr/installable-build.ts
+++ b/org/pr/installable-build.ts
@@ -8,9 +8,9 @@ const CIRCLECI_TOKEN: string = process.env['CIRCLECI_TOKEN']
 const PERIL_BOT_USER_ID: number = parseInt(process.env['PERIL_BOT_USER_ID'], 10)
 
 // This is a list of the CircleCI statuses to process
-const HOLD_CONTEXTS: string[] = ["ci/circleci: Installable Build/Hold", "ci/circleci: Installable Build/Approve WordPress"]
+const HOLD_CONTEXTS_COMMENT: string[] = ["ci/circleci: Installable Build/Hold", "ci/circleci: Installable Build/Approve WordPress"]
 // In some cases, we want to update the status of the job, but we don't want to post a CFA comment.
-const HOLD_CONTEXTS_NO_COMMENT: string[] = ["ci/circleci: Installable Build/Approve Jetpack"]
+const HOLD_CONTEXTS: string[] = ["ci/circleci: Installable Build/Approve Jetpack"]
 // The following statuses are described with regex in order to be able to handle special cases where the app name is added to the state, 
 // e.g. "ci/circleci: Jetpack Installable Build"
 const INSTALLABLE_BUILD_CONTEXTS: string[] = ["ci\/circleci:.*Installable Build$", "ci\/circleci:.*Test Android on Device$"] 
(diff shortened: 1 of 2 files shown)


org/pr/installable-build.ts
```diff
@@ -136,7 +138,7 @@ async function getDownloadCommentText(status) {
 }
 
 export default async (status: Status) => {
-    if (status.state == "pending" && HOLD_CONTEXTS.includes(status.context)) {
+    if (status.state == "pending" && HOLD_CONTEXTS.concat(HOLD_CONTEXTS_NO_COMMENT).includes(status.context)) {
       await markStatusAsSuccess(status)
       await createOrUpdateComment(status, `You can trigger an installable build for these changes by visiting CircleCI [here](${status.target_url}).`)
     } else if (status.state ==  "success" && INSTALLABLE_BUILD_CONTEXTS.filter(s => status.context.match(s)).length > 0) {
```
AliSoftware, line 144: If I'm not mistaken, this is the only place where both `HOLD_CONTEXTS` and `HOLD_CONTEXTS_NO_COMMENTS` constants are used.

 - If that's intended, and the only place where they are used here they are concatenated, not sure what's the point of having them split in the first place?
 - Instead I think what you intended to do is to NOT call `createOrUpdateComment` for the case of `HOLD_CONTEXTS_NO_COMMENTS`?

loremattei: Oh, I didn't notice my last push failed 🤦! I guess this is what you get when you try to write some code while you have to respond to too many pings 😂

loremattei: Yeah! That version was clearly doing nothing :-)

org/pr/installable-build.ts
```diff
@@ -140,7 +140,9 @@ async function getDownloadCommentText(status) {
 export default async (status: Status) => {
     if (status.state == "pending" && HOLD_CONTEXTS.concat(HOLD_CONTEXTS_NO_COMMENT).includes(status.context)) {
       await markStatusAsSuccess(status)
-      await createOrUpdateComment(status, `You can trigger an installable build for these changes by visiting CircleCI [here](${status.target_url}).`)
+      if (HOLD_CONTEXTS.includes(status.context)) {
```
AliSoftware, line 143: Wouldn't it be more exact and clearer to write it like this? 🤔 

```suggestion
      if (!(HOLD_CONTEXTS_NO_COMMENT.includes(status.context))) {
```

AliSoftware (commented):
Aside from stylistic comment, LGTM, but I'll let @jkmassel approve since he's the one knowing what the original intent was here.

org/pr/installable-build.ts
loremattei: I see 🤔 I'm not sure this is clearer or more exact. I mean: it makes sense for us because we know that, historically, there was a single flow and we are setting up an exception. 

But, in general, if you don't know or don't remember the history, you have two functions and you want to call the second one only for a smaller set of contexts, so the negated condition may end up being less clear.

AliSoftware: Right. My reasoning when I read the code was more about "since this guards the call to `createOrUpdateComment`, this makes sense to me that we'd only do it for context we specified by the "…_NO_COMMENT" constant." But I thought this also probably because the `HOLD_CONTEXTS` constant name doesn't tell anything about comment or not, while `HOLD_CONTEXTS_NO_COMMENT` does have the "comment" condition in its name 🤷 

Alternatively, renaming the first constant `HOLD_CONTEXTS_COMMENT` (and then use your positive condition `if (HOLD_CONTEXTS_COMMENT.includes(status.context))` would have the same effect (and might even be better to make it clear about the differences? )

Anyway, I'm ok with both tbh, it's not a big deal and just stylistic anyway, so I'll leave it up to you to chose what feels clearer 🙂

loremattei: Makes sense. I updated the name of the arrays.

PR #92, opened 2021-12-08
Add qrcode version of apk links to comments on wp android

enejb:
Currently when you see the Comment on PR that have the APK links it is hard to use them on the phone. 
You have to type them out or send the url over to your phone somehow. 

This PR makes it easier to just use the QRCode reader on your phone to get the links. 

## How has this been tested.

I update the test and made the test pass.


org/pr/installable-build.ts
```diff
@@ -131,7 +135,7 @@ async function getDownloadCommentText(status) {
   if (apkArtifacts.length == 1) {
     return `You can test the changes on this Pull Request by downloading the APK [here](${apkArtifacts[0].url}).`
   } else if (apkArtifacts.length > 1) {
-    const links = apkArtifacts.map(artifact => ` - [${artifact.path.split("/").pop()}](${artifact.url})`).join(`\n`)
+    const links = apkArtifacts.map(artifact => ` - [${artifact.path.split("/").pop()}](${artifact.url})\n<img src="${ generateQR( artifact.url )}" />`).join(`\n`)
```
AliSoftware, line 138: I'm late to the party (sorry I saw the review request but was too busy to get to it sooner), but what do you think (either in a next iteration… or more likely, when @jkmassel you'll re-implement Installable-Build Comments directly via buildkite pipeline) to use a collapsible section instead?

The odd thing with this current HTML is that it starts a bullet-point list but that list will contain large images (of the QR code) between the items. Not big a deal, but a bit unusual for bullet lists. But then there's also the fact that it can quickly take quite some vertical space in the comment, especially with 2 QR codes (one for WP, one for JP) one below the other.

So I'm suggesting something like this instead:

```suggestion
    const links = apkArtifacts.map(artifact => `<details><summary><a href="${artifact.url}">${artifact.path.split("/").pop()}</a></summary><img src="${generateQR(artifact.url)}" /></details>
```

Which should render something like below, where QR codes are hidden by default to reduce vertical space, but can be shown inline with a single click. WDYT?

---

You can test the changes on this Pull Request by downloading the APKs:
<details><summary><a href="https://117588-9306568-gh.circle-artifacts.com/0/Artifacts/Jetpack-pr-15683-build-117588.apk">Jetpack-pr-15683-build-117588.apk</a></summary></details>
<details><summary><a href="https://117588-9306568-gh.circle-artifacts.com/0/Artifacts/WordPress-pr-15683-build-117588.apk">WordPress-pr-15683-build-117588.apk</a></summary></details>

enejb:
Thanks for the feedback. @AliSoftware 
I really like that you can see the QRCodes right away it makes it easier for folks to realize that you can now test the PR by pointing to your phone and getting the latest changes. 
That being said there is definitely room for improvement in the way the markup is renderd. 

I was thinking about creating some sort of table view. With the QR codes on the same line. 

<table>
<tr><td>WordPress App</td><td>Jetpack App</td></tr>
<tr><td>
<a href="">  </a>
</td><td>
<a href=""></a> </td></tr>
</table>

So that thing are displayed nicer. What do you think?

AliSoftware:
I like that table idea a lot, and it looks indeed even better than my idea 👍

PR #95, opened 2022-03-14
Try to fix check-tracks.ts for diffForFile disappearance bug

AliSoftware:


org/pr/check-tracks.ts
```diff
@@ -25,29 +24,12 @@ async function checkTracksManagementFiles() {
...
-            if (danger !== undefined) {
-                console.log("Danger is no longer defined")
-            }
-            else {
-                console.log("Danger Git Object: ", danger.git)
-            }
-        }
-
         const diff = await git.diffForFile(thisFile);
         if (/AnalyticsTracker\.track/.test(diff.diff)) {
```
AliSoftware, line 34: See question I've left [in this WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/5964/files#r826237040): should we also check matches against `analyticsTrackerWrapper.track(…)` as well? \cc @shiki

AliSoftware: Ok after some chat in Slack on the other WCAndroid PR, let's keep that question for a separate brainstorming session and potential future PR, because:
1. the [variable pointing to the wrapper can be named whatever the developers want](https://github.com/woocommerce/woocommerce-android/pull/5964#discussion_r826242322) so there's no magic RegEx which could allow us to detect that in every context and use case possible
2. Even switching the RegEx to something more generic, like `\.track([A-Z_]` to cover both cases… might work for Android, where `track` can take either `track(AnalyticsEvent.USED_ANALYTICS)` or `track(USED_ANALYTICS)` equally in Kotlin… but that might not match how things are done in Swift, and we'll need to ensure that the RegEx applies to all codebases and all platforms

So this requires someone with more knowledge about the codebase around Tracks call sites and use cases and maybe some brainstorming to ensure we don't break the rule by changing the RegEx, which means we should keep that for a separate PR instead of trying to change it here in precipitation.

AliSoftware:
Confirmed in https://github.com/woocommerce/woocommerce-android/pull/5964 that this Peril PR fixed the issue (after this Peril PR being merged, then retriggering Peril in the WCAndroid PR making Peril not crash anymore)

PR #102, opened 2022-09-07
Decrease Woo PR Milestone Warning days

toupper:
# Description
Since we recently moved to 1-week release cycles and milestones matching these releases, we always get those warnings if we assign the PR to the next milestone (which happens often) because it is always closing in less than 4 days. That adds too much noise and loses the original intent of those warnings. Because of that, we decrease the warning days to two with this PR.

> Internal Ref: pdnsEh-B2-p2

# Changes
- Create a new milestone file for woo with 2 warning days
- Point to that file in the peril settings json for woo ios and android

### DIFF
--- a/org/pr/milestone-woo.ts
+++ b/org/pr/milestone-woo.ts
@@ -1,0 +1,53 @@
+import {warn, danger} from "danger";
+
+export default async () => {
+    // Create consts for everything we need from `danger` object as soon as possible
+    // so it's not dropped during await operation
+    const githubLabels = danger.github.issue.labels;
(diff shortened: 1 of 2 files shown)


mokagio (approved):
```
➜ diff org/pr/milestone.ts org/pr/milestone-woo.ts
36,37c36,37
<     // Warn if the milestone is closing in less than 4 days
<     const warningDays = 4;
---
>     // Warn if the milestone is closing in less than 2 days
>     const warningDays = 2;
44c44
<
---
>
```

PR #109, opened 2023-02-03
Allow Ruby post-1.9 Hash syntax in `Podfile` (`commit:`)

mokagio:
I plan to update the Peril rules to validate the `Podfile` to:

- Allow for using `branch: 'trunk'`, but fail if the branch name is different
- Warn when using a `-beta` version, either in a version rule `~> 1.0-beta` or in a tag `tag: '1.0.0-beta`

Before doing so, I took some time to extract the `Podfile` tests in a dedicated file (wasn't sure about extracting the logic in a dedicated function, though) so they're easier to find and won't clutter the rest of the code.


org/pr/ios-macos.ts
```diff
@@ -16,8 +16,9 @@ export default async () => {
 
     // Podfile should not reference commit hashes
     const podfileContents = await danger.github.utils.fileContents("Podfile");
-    const matches = podfileContents.match(/^[^#[]*:commit/gm);
-    if (matches !== null) {
+    const matchesOldRuby = podfileContents.match(/^[^#[]*:commit =>/gm);
+    const matchesNewRuby = podfileContents.match(/^[^#[]*commit:/gm);
```
AliSoftware, line 20: What is the point of excluding the `[` in `[^#[]`?

```suggestion
    const matchesOldRuby = podfileContents.match(/^[^#]*:commit =>/gm);
    const matchesNewRuby = podfileContents.match(/^[^#]*commit:/gm);
```

AliSoftware: Also I wonder if there's a risk of people using more spaces between `:commit` and `=>`, maybe a ` *` would be better?

Which makes me wonder… should we really lint the `Podfile` itself? Wouldn't it be more resilient to lint the `Podfile.lock` (which can be parsed as a `YAML` file) and detect pod referenced as commits from there, not depending on the syntax used by the user but instead relying on machine-generated counterpart?

crazytonyli: We can potentially use `be pod ipc podfile-json Podfile` for easier parsing?

mokagio: @AliSoftware:

> Which makes me wonder… should we really lint the `Podfile` itself? Wouldn't it be more resilient to lint the `Podfile.lock` (which can be parsed as a YAML file)

Ah, great idea. I went in and sort of patched the existing logic, but taking a step back reveals good improvements (good example of _local vs global maxima_).

I'll see what I can do on that front. The good thing is that the tests shouldn't need to change 😄

mokagio: @crazytonyli 

> We can potentially use `be pod ipc podfile-json Podfile` for easier parsing?

(I assume `be` is an alias for `bundle exec`–I use that, too)

I guess we could? However, Peril runs on Node and I'm tempted to try parsing the `Podfile.lock` YAML first, before trying to (I'm guessing) shell-out to run a Ruby command.

crazytonyli: > (I assume be is an alias for bundle exec–I use that, too)

Haha, yes indeed.

> Peril runs on Node

Oh, of course, sorry I thought this script runs on macOS.

AliSoftware: > I'm tempted to try parsing the `Podfile.lock` YAML first

See also https://github.com/wordpress-mobile/release-toolkit/issues/216 (where I left an example of how to do such parsing of pods list from the lockfile in Ruby)

org/pr/ios-macos.ts
```diff
@@ -14,11 +14,26 @@ export default async () => {
...
+    const podfileLockContents = await danger.github.utils.fileContents("Podfile.lock");
+    const podfileLockYAML = require("js-yaml").safeLoad(podfileLockContents);
+
+    if (podfileLockYAML != null && podfileLockYAML["DEPENDENCIES"] != null) {
+        // check if any pods are referenced from a commit hash
+        const podsReferencedByCommit = podfileLockYAML["DEPENDENCIES"].filter((pod: any) => pod.includes(", commit `"));
+        if (podsReferencedByCommit.length > 0) {
+            fail("Podfile: reference to a commit hash");
+        }
     }
```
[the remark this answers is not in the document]
AliSoftware: I've attempted an implementation of this in `alisoftware/extract-podfile-checks`, feel free to borrow it and land it in your branch if it works for you.

AliSoftware (changes requested):
Current implementation will fail on a `Podfile.lock` that contains transitive dependencies.

I've pushed a suggested improvement/fix [in `alisoftware/extract-podfile-checks`](https://github.com/Automattic/peril-settings/compare/alisoftware/extract-podfile-checks...mokagio/extract-podfile-checks) if you want to give that solution a try.

org/pr/ios-macos.ts
```diff
@@ -14,11 +14,26 @@ export default async () => {
...
+    //
+    // Verify by parsing Podfile.lock because it uses a standard format, unlike Podfile which might be written in different ways.
+    //
+    // Example of relevant Podfile.lock portion:
+    //
+    // DEPENDENCIES:
+    //     - Kanvas(from `https://github.com/tumblr/Kanvas-iOS.git`, branch `main`)
+    //     - WordPress - Editor - iOS(~> 1.19.8)
+    //     - WordPressUI(from `https://github.com/wordpress-mobile/WordPressUI-iOS`, commit `5ab5fd3dc8f50a27181cf14e101abe3599398
+// cad`)
```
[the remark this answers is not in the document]
mokagio: Ahah. That's interesting as I used VS Code for this PR, maybe it has some settings I'm not aware of. Good catch.

mokagio:
BTW, I can now see the CI pages. Thanks for the group wrangling @AliSoftware

org/pr/ios-macos.ts
```diff
@@ -1,10 +1,27 @@
 import {message, warn, fail, danger} from "danger";
 
+// Example of relevant Podfile.lock portion:
+//
+// DEPENDENCIES:
+//     - Kanvas(from `https://github.com/tumblr/Kanvas-iOS.git`, branch `main`)
+//     - WordPress - Editor - iOS(~> 1.19.8)
+//     - WordPressUI(from `https://github.com/wordpress-mobile/WordPressUI-iOS`, commit `5ab5fd3dc8f50a27181cf14e101abe3599398cad`)
+function parseCommitPods(list: string[], entry: any): string[] {
```
AliSoftware, line 9:
```suggestion
// Function used as a reducer to parse and accumulate pods which are referenced by commit.
// Uses recursion to parse transitive dependencies of each entry.
//
// @param list - The accumulated list of commit-referenced pods so far
// @param entry - an entry representing a pod listed in the lockfile.
//                Will be a `string` if the pod has no dependencies.
//                Otherwise, will be an `object`, with the pod being the single key, and the value being the dependencies
// @return The new list with any commit-referenced pod found added to the initial `list`
function parseCommitPods(list: string[], entry: any): string[] {
// Example of relevant Podfile.lock portion:
//
// DEPENDENCIES:
//     - Kanvas (from `https://github.com/tumblr/Kanvas-iOS.git`, branch `main`)
//     - WordPress-Editor-iOS (~> 1.19.8)
//     - WordPressUI (from `https://github.com/wordpress-mobile/WordPressUI-iOS`, commit `5ab5fd3dc8f50a27181cf14e101abe3599398cad`)
//     - WordPressKit (~> 6.1.0-beta):
//      - WordPressShared (from `https://github.com/test/pod2.git`, commit `e123bd0a9fef58a5897ed2101044f56a42e614c7`)
```

tests/ios-podfile.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    dm.danger.github.utils.fileContents.mockReturnValue(Promise.resolve(""));
+});
+
+describe("Podfile should not reference commit hashes checks", () => {
+    it("fails when finds a commit reference", async () => {
+        dm.danger.github.utils.fileContents.mockReturnValueOnce(
+            Promise.resolve(
+                `DEPENDENCIES:
+                    - Gutenberg (from \`https://github.com/wordpress-mobile/gutenberg-mobile.git\`, tag \`v1.88.0\`)
+                    - TestPod(from \`https://github.com/test/pod.git\`, commit \`82f65c9cac2b59940db219a2327e12a0adfea4e3\`)
```
AliSoftware, line 46: [I just checked and the actual syntax in `Podfile.lock` has a space between the pod name and open parenthesis](https://github.com/wordpress-mobile/wordPress-iOS/commit/ab2cf5b07ce68947f781f7f9b1a28e80cc5fc73f#diff-e5b768ab2330ee29074f2eeadddd406f19e73b14673eb5f9b5ed6b5b77f4e363L557-R557) (as one would expect)

```suggestion
                    - TestPod (from \`https://github.com/test/pod.git\`, commit \`82f65c9cac2b59940db219a2327e12a0adfea4e3\`)
```

org/pr/ios-macos.ts
```diff
@@ -1,10 +1,27 @@
...
+    if (typeof entry === 'string') {
+        const match = entry.match(/(.*)\(from .*, commit `.*`/);
```
AliSoftware, line 11:
```suggestion
        const match = entry.match(/(.*) \(from .*, commit `.*`/);
```

tests/ios-podfile.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    })
+
+    it("fails when finds commit references in transitive dependencies", async () => {
+        dm.danger.github.utils.fileContents.mockReturnValueOnce(
+            Promise.resolve(
+                `DEPENDENCIES:
+                    - WordPressKit (~> 6.1.0-beta)
+                    - TestPod (~> 1.7.2):
+                      - TestDep1 (~> 2.0-beta)
+                      - TestDep2(from \`https://github.com/test/pod2.git\`, commit \`5dbb1b2ef4b3b8157569df5878b9ea67e3a9377a\`)
```
AliSoftware, line 64:
```suggestion
                      - TestDep2 (from \`https://github.com/test/pod2.git\`, commit \`5dbb1b2ef4b3b8157569df5878b9ea67e3a9377a\`)
```

tests/ios-podfile.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            Promise.resolve(
+                `DEPENDENCIES:
+                    - Gutenberg (from \`https://github.com/wordpress-mobile/gutenberg-mobile.git\`, tag \`v1.88.0\`)
+                    - TestPod(from \`https://github.com/test/pod.git\`, commit \`82f65c9cac2b59940db219a2327e12a0adfea4e3\`):
+                      - TestDep1 (~> 2.0-beta)
+                      - TestDep2(from \`https://github.com/test/pod2.git\`, commit \`5dbb1b2ef4b3b8157569df5878b9ea67e3a9377a\`)
+                      - TestDep3 (~> 1.7.2)
+                    - WordPressKit (~> 6.1.0-beta):
+                      - WordPressShared(from \`https://github.com/test/pod2.git\`, commit \`e123bd0a9fef58a5897ed2101044f56a42e614c7\`)
+                    - StandalonePod (~> 1.2.7)
```
AliSoftware, line 86:
```suggestion
                    - Gutenberg (from \`https://github.com/wordpress-mobile/gutenberg-mobile.git\`, tag \`v1.88.0\`)
                    - TestPod (from \`https://github.com/test/pod.git\`, commit \`82f65c9cac2b59940db219a2327e12a0adfea4e3\`):
                      - TestDep1 (~> 2.0-beta)
                      - TestDep2 (from \`https://github.com/test/pod2.git\`, commit \`5dbb1b2ef4b3b8157569df5878b9ea67e3a9377a\`)
                      - TestDep3 (~> 1.7.2)
                    - WordPressKit (~> 6.1.0-beta):
                      - WordPressShared (from \`https://github.com/test/pod2.git\`, commit \`e123bd0a9fef58a5897ed2101044f56a42e614c7\`)
                    - StandalonePod (~> 1.2.7)
```

AliSoftware (commented):
Suggested adding some TSDoc to document `parseCommitPods`, but also we need to fix the tests and regex to account for the fact that in the real case there is a space between the pod name and the parentheses (otherwise the tests will pass… but would not reflect real world cases)

org/pr/ios-macos.ts
mokagio: I won't commit this suggestion because I think the indentation of the comment inside the function body got mangled by the `suggestion` block. It's inline with the method definition, but should be inline with the content instead.

org/pr/ios-macos.ts
```diff
@@ -44,3 +49,30 @@ export default async () => {
...
+// @return The new list with any commit-referenced pod found added to the initial `list`
+function parseCommitPods(list: string[], entry: any): string[] {
+    // Example of relevant Podfile.lock portion:
+    //
+    // DEPENDENCIES:
+    //     - Kanvas (from `https://github.com/tumblr/Kanvas-iOS.git`, branch `main`)
+    //     - WordPress-Editor-iOS (~> 1.19.8)
+    //     - WordPressUI (from `https://github.com/wordpress-mobile/WordPressUI-iOS`, commit `5ab5fd3dc8f50a27181cf14e101abe3599398cad`)
+    //     - WordPressKit (~> 6.1.0-beta):
+    //       - WordPressShared (from `https://github.com/test/pod2.git`, commit `e123bd0a9fef58a5897ed2101044f56a42e614c7`)
```
AliSoftware, line 69: Nit:
```suggestion
    //       - WordPressShared (from `https://github.com/wordpress-mobile/WordPress-iOS-Shared.git`, commit `e123bd0a9fef58a5897ed2101044f56a42e614c7`)
```

tests/ios-podfile.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        dm.danger.github.utils.fileContents.mockReturnValueOnce(
+            Promise.resolve(
+                `DEPENDENCIES:
+                    - Gutenberg (from \`https://github.com/wordpress-mobile/gutenberg-mobile.git\`, tag \`v1.88.0\`)
+                    - TestPod (from \`https://github.com/test/pod.git\`, commit \`82f65c9cac2b59940db219a2327e12a0adfea4e3\`):
+                      - TestDep1 (~> 2.0-beta)
+                      - TestDep2 (from \`https://github.com/test/pod2.git\`, commit \`5dbb1b2ef4b3b8157569df5878b9ea67e3a9377a\`)
+                      - TestDep3 (~> 1.7.2)
+                    - WordPressKit (~> 6.1.0-beta):
+                      - WordPressShared (from \`https://github.com/test/pod2.git\`, commit \`e123bd0a9fef58a5897ed2101044f56a42e614c7\`)
```
AliSoftware, line 85:
```suggestion
                      - WordPressShared (from \`https://github.com/wordpress-mobile/WordPress-iOS-Shared.git\`, commit \`e123bd0a9fef58a5897ed2101044f56a42e614c7\`)
```

AliSoftware (approved):
Nitpicking to adjust URLs in examples/tests to be more consistent/realistic, but apart from that, LGTM 👍

PR #110, opened 2023-02-14
Support nested dependencies from `Podfile.lock`

mokagio:
This PR tracks @AliSoftware's suggestions on top of #109.

GitHub reports the tests failing via its Checks, but when I select the [build](https://buildkite.com/automattic/peril-settings/builds/45), it 404s.

image

Interestingly, Buildkite can't find a `peril-settings` pipeline:

image

🤔 

[There's been an incident a couple of hours ago](https://www.buildkitestatus.com/incidents/t7hn25fb7fds), which _might_ have to do with this although it should have been resolved by now. I'll wait a bit and check again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,19 @@
 
 These are the settings for running [Peril](https://github.com/danger/peril) on repositories in `Automattic`, `simplenote` and `wordpress-mobile`.
 
-### Installing
+## Development Environment
 
-After cloning the repo, run `yarn install`.
+### Using Docker
 
-### Running the tests
+You can use Docker to work and test this project in the appropriate container.
 
-Tests can be run with `yarn test`.
+```
+docker run -it --rm --init --volume "${PWD}":/workdir --workdir /workdir node:10 /bin/sh -e -c $'yarn install --frozen-lockfile; yarn jest'
+```
+
+### Without Docker
+
+ - Clone the repo.
+ - Ensure you have `node` and `yarn` installed on your machine.
+ - Run `yarn install`.
+ - Run the tests using `yarn jest`.
--- a/org/pr/ios-macos.ts
+++ b/org/pr/ios-macos.ts
@@ -1,15 +1,20 @@
 import {message, warn, fail, danger} from "danger";
 
-function findCommitPods(entry: any): string[] {
+// Example of relevant Podfile.lock portion:
+//
+// DEPENDENCIES:
+//     - Kanvas(from `https://github.com/tumblr/Kanvas-iOS.git`, branch `main`)
+//     - WordPress - Editor - iOS(~> 1.19.8)
+//     - WordPressUI(from `https://github.com/wordpress-mobile/WordPressUI-iOS`, commit `5ab5fd3dc8f50a27181cf14e101abe3599398cad`)
+function parseCommitPods(list: string[], entry: any): string[] {
     if (typeof entry === 'string') {
         const match = entry.match(/(.*)\(from .*, commit `.*`/);
-        return (match != null) ? [match[1]] : [];
+        if (match != null) { list.push(match[1]) };
     } else {
         const key: string = Object.keys(entry)[0];
-        const keyMatch: string[] = findCommitPods(key);
-        const depsMatches: string[] = entry[key].flatMap(findCommitPods);
-        return keyMatch.concat(depsMatches);
+        list = [key, ...entry[key]].reduce(parseCommitPods, list);
     }
+    return list
 }
 
 export default async () => {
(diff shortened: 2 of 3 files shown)


AliSoftware:
> GitHub reports the tests failing via its Checks, but when I select the build, it 404s.

I can see https://buildkite.com/automattic/peril-settings/builds/45

mokagio:
@AliSoftware 

> > GitHub reports the tests failing via its Checks, but when I select the build, it 404s.

> I can see [buildkite.com/automattic/peril-settings/builds/45](https://buildkite.com/automattic/peril-settings/builds/45)

I see your comment only now. In the meantime, I realized this is most likely an issue with the access my user has, in particular because #111 triggered a new build, number 46 which is consistent with the latest one here being 45, and I can't access that either. I also can't access previous builds on `trunk`.

I must no longer be part of the team that has access to that pipeline on Buildkite.

AliSoftware:
> I must no longer be part of the team that has access to that pipeline on Buildkite.

@mokagio  I just added you back to the relevant teams on Buildkite, you should be able to see the builds now.

AliSoftware:
@mokagio Ok, so I found the source of the issue: it's because `Array.flatMap` is not available in `node:10`, and only started to be available in `node:11`.

That investigation led me to some rabbit hole, where I tried to updated some dependencies, especially:
 - Update the TypeScript compiler from `es2017` to `es2019` (as VSCode was suggesting me to). ❌  This didn't solve the issue 
 - Then update our pipeline to run in a `node:16` container instead of `node:10`. ✅  This _did_ solve the issue for `flatMap` not existing in `node:10` and made the CI go green
 - But then saw that [there were some scary warnings about old dependencies not being maintained anymore](https://github.com/Automattic/peril-settings/pull/110/commits/2053b7eb67346ed1f1a6db30fc7026efb9c26583), so decided to upgrade `danger` to version `11.x` so that it also used the more recent dependencies without those warnings. But that made the CI go red again, not too unsurprisingly because `danger ^11.0.0` is a breaking change compared to `danger ^10.0.0`, so we'd have to update the codebase accordingly

But then I realized that we are using Peril, not Danger. And the Peril server [still uses `node:10`](https://github.com/danger/peril/blob/master/api/package.json#L33-L35) and apparently even [depends on `danger: 9.1.3`](https://github.com/danger/peril/blob/master/api/package.json#L86)?! (though [their runner depends on `7.0.19`](https://github.com/danger/peril/blob/master/runner/package.json#L38)… so now I'm quite confused… but [according to the diff from that PR](https://github.com/danger/peril/pull/449/files) I think the version of danvger in `api/package.json` is the one that counts… which means we should have wrote our own dependency in this repo as `danger ^9.0.0` in the first place, not `danger ^10.0.0` and it's only still working in danger 10 by sheer luck?).

In any case, I'm not entirely sure how this whole thing work (not being a npm/yarn/node/typescript expert in the first place), and a bit confused if it'd be ok to upgrade node and danger in this setup without breaking our instance of the Peril server running those even if it doesn't break our CI (which only runs the danger module itself, not via Peril)

---

TL;DR: Maybe it wasn't a good idea to update to `node:16` in the first place—especially if it means we'll be required to update to `danger ^11.0.0` to get rid of the big warning on old dependencies—and we'll have to stick with `node:10`.
Which means that we'll have to find an alternative solution to using `flatMap` to implement the recursion I've added in that PR's diff to handle transitive dependencies…

AliSoftware:
Reworked the code to [use a reducer](https://github.com/Automattic/peril-settings/pull/110/commits/e7724d9954ec9c8d9745506c853ca270d56a608c) (`reduce` and a `parseCommitPods(accumulator, element) -> newList` reducer), which now makes the CI pass 🎉 

I've also updated the `README.md` instructions to suggest to use `docker` to run `yarn install && yarn jest`, so that it ensures that we use the `node:10` docker image and node version, i.e. the same as CI and the same as Peril, instead of relying to the locally-installed node version like I was doing which explained why the tests passed locally on my Mac (I have `node 16` installed, apparently) but not on CI.

PS: I've kept my attempt at updating `node:16` and `danger:^11` (the one I was talking about in my previous comment above) in the `alisoftware/update-danger` branch if you are curious and wanted to take a look.

mokagio (commented):
Thanks @AliSoftware !

This looks great. For some reason, my approve button is disabled, but this should be considered as an approval. I'll see if I can merge it myself next, otherwise, please do so.

AliSoftware:
> For some reason, my approve button is disabled

FYI, this is because GitHub (rightfully) does not let you approve your own PRs 😉

PR #117, opened 2023-11-20
Remove checks from WordPress iOS that now run on Danger

mokagio:
The `Dangerfile` and dangermattic version I referred to for these changes was from b2a5d65112d6580c29c8b7f5ab33b00c35a0759d, https://github.com/wordpress-mobile/WordPress-iOS/tree/b2a5d65112d6580c29c8b7f5ab33b00c35a0759d

Notice I didn't remove the rules to run the `ios-macos.ts` checks because I noticed those checks include one for the Core Data model, which I couldn't find replicated either in dangermattic or in the local WordPress iOS setup.

### DIFF
--- a/peril-settings.json
+++ b/peril-settings.json
@@ -12,18 +12,10 @@
                 "Automattic/peril-settings@org/pr/ios-macos.ts"
             ],
             "pull_request": [
-                "Automattic/peril-settings@org/pr/ios-macos.ts",
-                "Automattic/peril-settings@org/pr/milestone.ts"
-            ],
-            "pull_request.opened, pull_request.labeled, pull_request.unlabeled": [
-                "Automattic/peril-settings@org/pr/label.ts"
+                "Automattic/peril-settings@org/pr/ios-macos.ts"
             ],
             "issues.opened, issues.labeled, issues.unlabeled": [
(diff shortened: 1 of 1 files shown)


iangmaia:
> The `Dangerfile` and dangermattic version I referred to for these changes was from b2a5d65112d6580c29c8b7f5ab33b00c35a0759d, https://github.com/wordpress-mobile/WordPress-iOS/tree/b2a5d65112d6580c29c8b7f5ab33b00c35a0759d

Thanks! I was refraining for now to remove these checks already as we still need to have Danger run on label, milestone and other PR changes. I don't mind if we do it now but we need to be clear with what we'll be missing (hopefully, not for long 🙂 ).

> Notice I didn't remove the rules to run the `ios-macos.ts` checks because I noticed those checks include one for the Core Data model, which I couldn't find replicated either in dangermattic or in the local WordPress iOS setup.

Those are implemented here: https://github.com/Automattic/dangermattic/pull/9
It's the last remaining pending PR. As soon as it's merged, I'll add calls to it together with other improvements.

> @iangmaia I couldn't ask you for a review directly because your user was not part of the team that can review this repo. I rectified that and you should have an invite in your inbox. This ping is to avoid having to wait to know the invitation has been accepted to ask for a review.

Thank you!

PR #118, opened 2023-11-23
Update iOS checks to prevent pods to be referenced by branch, not commit

mokagio:
The same as https://github.com/Automattic/dangermattic/pull/30 but implemented here in Peril just so we don't get conflicting comments while Peril still runs on PRs.

I'm very happy we had unit tests for this, because I think the only way to verify its behavior otherwise is to merge the change and see how it works in production.


iangmaia (approved):
I was wondering if it was worth it to add this here at all, but since it's already done... 🙂
Good for some transition time as well until it is disabled completely 👍

